PR #2, opened 2016-01-19
Clone config when inheriting

me:
When a `Configurable` is subclassed, if the config of the subclass gets modified, it will overwrite the parent's config. This fixes the issue.

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -40,7 +40,7 @@ module Dry
     def inherited(subclass)
       subclass.instance_variable_set(:@_config_mutex, @_config_mutex)
       subclass.instance_variable_set(:@_settings, @_settings)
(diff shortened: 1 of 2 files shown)


AMHOL:
Thanks for the PR, I think we should `dup`/`clone` the `@_settings` ivar too in this scenario, and probably create a new mutex as the state is no longer shared. Let me know what you think.

me:
Good point. Updated, what do you think?

AMHOL:
Looks good, thanks :+1:

AMHOL:
I will release tonight, you OK to pull from the master branch in the meantime if this is causing you issues?

me:
Thanks! That's completely fine. My issues were in `dry-validation` actually, and I was already using master to get the inherited config :) So it will be great to have the new version out.

solnic:
oh, damn, I thought I've fixed that already here as I noticed that issue as well :(

PR #3, opened 2016-03-23
switch to using concurrent instead of thread_safe

joevandyk:


AMHOL:
Thanks @joevandyk :smile:

PR #5, opened 2016-06-22
Synchronize access to @_config only during write

splattael:
Prior this commit read and write access to `#config` is synchronized.

This commit removes the need of synchronizing when reading an already
initialized @_config.

Writing @_config (typically during boot time) via `configure { ... }`
is still synchronized which "feels safe" (TM).

What do you think? Is "feels safe" good enough for us? :-)

Big pro: Performance improvement!

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -48,9 +48,9 @@ module Dry
     #
     # @api public
     def config
+      return @_config if defined?(@_config)
       @_config_mutex.synchronize do
-        return @_config if defined?(@_config)
-        @_config = Config.new(*_settings.keys).new(*_settings.values) unless _settings.empty?
+        @_config ||= Config.new(*_settings.keys).new(*_settings.values) unless _settings.empty?
       end
     end
 


lib/dry/configurable.rb
```diff
@@ -48,8 +48,8 @@ def inherited(subclass)
     #
     # @api public
     def config
+      return @_config if defined?(@_config)
       @_config_mutex.synchronize do
-        return @_config if defined?(@_config)
```
flash-gordon: Good catch, but you can't just remove this line, the result won't be equivalent, consider this:
thread 1
@_config not defined, move on
acquire lock

thread 2
@_config not defined, move on
wait on mutex

thread 1
set @_config
release mutex

thread 2
acquire mutex
set @_config <- NOT SAFE
release mutex

So, along your return statement you should change `@_config = ` with `@_config ||=`

splattael: That's true. I'll change it! Thanks :-)

splattael:
@flash-gordon Done!

AMHOL:
Not sure about this myself, although I guess it doesn't really matter as configuration should be handled before you get to the request phase. Do you have any measurable results on the performance improvement?

splattael:
@AMHOL @solnic and me were benchmarking and profiling `dry-validation` and `Configurable#confing` popped up as it has a lot of read access.

I'll provide some soonish! :-)

splattael:
Hi,

we used `benchmarks/benchmark_form_invalid.rb` in `dry-validation`. Sadly, I cannot reproduce my results in `dry-v` :-(

Benchmarking `Configurable#config` alone still shows a performance boost:

```ruby
require 'benchmark/ips'
require 'dry-configurable'

class Testing
  extend Dry::Configurable

  setting :foo
end

Testing.config

Benchmark.ips do |x|
  x.report "read access" do
    Testing.config
  end
end
```

## Before

```
Warming up --------------------------------------
         read access   125.527k i/100ms
         read access      2.092M (± 0.7%) i/s -     10.544M in   5.039468s
```

## After

```
Warming up --------------------------------------
         read access   192.583k i/100ms
         read access      5.297M (± 2.5%) i/s -     26.576M in   5.020656s
```

WDYT?

Kind regards,
  Peter

splattael:
The numbers for jruby (9.1.2.0) are even more impressive :flushed: 

## Before

```
Warming up --------------------------------------
         read access    40.740k i/100ms
         read access    937.474k (± 8.5%) i/s -      4.644M in   4.998761s
```

### After

```
Warming up --------------------------------------
         read access   127.524k i/100ms
         read access      6.471M (± 9.6%) i/s -     31.881M in   4.991136s
```

AMHOL:
Yeah, I doubt it will make much of a difference in libraries as config shouldn't be read that much really, merged nevertheless

splattael:
@AMHOL Thanks for merging.
You are right, reading config needs be reduced in `dry-v`!

PR #10, opened 2016-08-30
Add lazily evaluated settings

AMHOL:
This adds lazily evaluated default settings, with the following API:

```ruby
class App
  extend Dry::Configurable

  setting :db_adapter, 'postgres'
  setting :db_user, 'root'
  setting :db_password, ''
  setting :db_host, '127.0.0.1'
  setting :db_port, '127.0.0.1'
  setting :db_name, 'development'
  # This \/ \/
  setting :dsn do |config|
    "#{config.db_adapter}://#{config.db_user}:#{config.db_password}@#{config.db_host}:#{config.db_port}/#{config.db_name}"
  end
end
```

Passing a block with an arity of 0 will still create a nested config as it did before, note that this implementation does not give nested configs access to parent configs, i.e.

```ruby
class App
  extend Dry::Configurable

  setting :database do
    setting :adapter, 'postgres'
    setting :user, 'root'
    setting :password, ''
    setting :host, '127.0.0.1'
    setting :port, '127.0.0.1'
    setting :name, 'development'
    # This only has access to "sibling" config values, i.e.
    # adapter, user, password, host, port and name
    # it does not have access to any other configs
    setting :dsn do |config|
      "#{config.db_adapter}://#{config.db_user}:#{config.db_password}@#{config.db_host}:#{config.db_port}/#{config.db_name}"
    end
  end

  setting :other_stuff do
    setting :blah
  end
end
```


jasdeepsingh:
This is not working for me, Return nil always. Here's my code:

```ruby
class App
  extend Dry::Configurable

  setting :api_username
  setting :api_password
  setting :merchant_account
  setting(:environment, ActiveSupport::StringInquirer.new('test')) { |value| ActiveSupport::StringInquirer.new(value.to_s) }
  setting :endpoint do |config|
    "value"
  end
end
```

```
#=> App.config.endpoint 
#=> nil
```

What might be going on wrong here? I'm on `0.7.0` of gem.

PR #11, opened 2016-08-31
Add processor option to setting

AMHOL:
```ruby
class Foo
  extend Dry::Configurable

  setting :path, 'lib', processor: ->(value) { Pathname(value) }
end

Foo.config.path
# => #<Pathname:lib>

Foo.configure do |config|
  config.path = 'app'
end

Foo.config.path
# => #<Pathname:app>
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ rvm:
   - 2.1
   - 2.2
   - rbx-2
-  - jruby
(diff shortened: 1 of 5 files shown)


timriley:
I like it 👍

solnic:
Looks good

AMHOL:
Cheers @solnic @timriley

PR #16, opened 2016-09-07
Support value processor using a block argument with parameters passed to #setting

AMHOL:


AMHOL:
@solnic @timriley had to go with the block with parameter option for this, let me know what you think

AMHOL:
dry-v specs are passing with this

timriley:
Thanks for working on this @AMHOL, I think it'll be a good feature to have.

It's clever how you managed to get the `setting` method's block to be used in these 2 ways based on whether you pass it an arg, but I wonder if using this same mechanism for 2 different things is actually making the API less clear to use?

Is e.g. this:

```ruby
setting :foo, processor: -> value { Pathname(value) }
```

clearer than this?

```ruby
setting(:foo) { |value| Pathname(value) }
```

AMHOL:
@timriley I originally implemented this using a kwarg (https://github.com/dry-rb/dry-configurable/pull/11) which caused a bug and 0.2.0 had to be yanked

timriley:
Ah yes! I recall now :) Also, I was thinking about this a bit more and I realized there was an added benefit to using the block, in that you can only ever choose between one or the other: a processor _or_ nested settings, never both (which is a combination that doesn't make sense).

So yeah, looks good 👍🏼

Issue #9, opened 2016-08-15
Option parser


solnic:
Sometimes you may want to pre-process a value, ie:

``` ruby
class Foo
  extend Dry::Configurable

  setting :path { |value| Pathname(value) }
end
```

timriley:
This would be handy. In the past I've had to override it by doing something like this:

```ruby
class Foo
  extend Dry::Configurable
  setting :path

  def self.path
    Pathname(config.path)
  end
end
```

And then remembering to use that reader everywhere rather than the config value directly.

AMHOL:
@solnic @timriley this could be covered by https://github.com/dry-rb/dry-configurable/pull/10

```ruby
class Foo
  extend Dry::Configurable

  setting :path
  setting(:pathname) { |config| Pathname(config.path) }
end

Foo.configure do |config|
  config.path = 'test'
end

Foo.config.pathname
=> #<Pathname:test>
```

solnic:
Not sure if I follow, this seems like two different things

timriley:
Yeah, the fact that we end up with 2 configs in that example, when we really only wanted one that we could modify when it's set, shows that #10 (if we did decided to incorporate it) doesn't achieve the feature described here all that well?

AMHOL:
Yeah I was thinking about this the other night and came to the same conclusion

AMHOL:
@solnic @timriley WDYT about an interface like:

```ruby
class Foo
  extend Dry::Configurable

  setting :path, processor: ->(value) { Pathname(value) }
end
```

AMHOL:
See: https://github.com/dry-rb/dry-configurable/pull/11

PR #18, opened 2016-10-18
Add processor example to README

waiting-for-dev:
Just adds to the README `processor` option example, discussed in #9 and implemented in #11.

By the way, is there any prevision where a new version will be pushed to rubygems with this feature?

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ class App
   setting :adapter
   
   # Pre-process values
-  setting :path, 'test', processor: ->(value) { Pathname(value) }
+  setting(:path, 'test') { |value| Pathname(value) }
 end
 
 App.configure do |config|
--- a/README.md
+++ b/README.md
@@ -23,8 +23,12 @@ class App
     # Can pass a default value
     setting :dsn, 'sqlite:memory'
   end
+  
   # Defaults to nil if no default value is given
   setting :adapter
+  
+  # Pre-process values
+  setting :path, 'test', processor: ->(value) { Pathname(value) }
 end
 
 App.configure do |config|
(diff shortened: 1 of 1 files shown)


timriley:
Good idea to document this! I think, however, that you're looking at an out-of-date PR – this feature saw it's final implementation using a different API in #16.

Also, we're looking to move most of our documentation over to dry-rb.org instead of keeping them in READMEs. Would you mind submitting a PR to the [dry-rb.org repo](http://github.com/dry-rb/dry-rb.org) instead?

timriley:
We'll obviously need to wait until a release before the PR to dry-rb.org docs is merged - I'll let @AMHOL chime in here if he has any thoughts on release timing.

waiting-for-dev:
Thanks for pointing me to the updated API, I didn't saw that.

I submitted the pull request to dry-rb.org repository and also updated this one, because I think it is also useful if you find it in the Readme.

Cheers!

AMHOL:
Cheers, I know @solnic isn't keen on having documentation in the readme, I like it myself but I think it's better to be consistent so I will probably remove it here and in dry-container, I'll release tomorrow night, thanks for the reminder.

waiting-for-dev:
Ok, thinking about it a little bit I see @solnic is right. Having also a website, it is better to only have the documentation in one place, otherwise it may be difficult to keep both updated. I guess the README should only have a link to the website. I'll close this pull request, then. Thank you!

AMHOL:
@waiting-for-dev released as 0.3.0, sorry for the delay.

waiting-for-dev:
Hey, many thanks! :)

Issue #19, opened 2016-10-26
Reset configuration


waiting-for-dev:
I'm developing a gem that uses dry-configurable. In order to test it, I need to change some default settings in `before` hooks. I found that changed settings persist in subsequent tests. For that reason, I think I would need some `reset` method in order to apply it in `after` hooks.

What do you think about it? Would it be difficult to implement or does it have any drawback?

mereghost:
@waiting-for-dev A very hackish way to do that currently, is by removing `@_config` like this

```ruby
class App
  extend Dry::Configurable

  setting :important, 'default'

 def self.reset
  remove_instance_variable :@_config
 end
end
```

solnic:
We can add it as a test-specific interface that people could include on demand

PR #22, opened 2016-12-12
Allow access to settings through a hash interface

waiting-for-dev:
This allows using `#[]` and `#[]=` to get/set settings in the
configuration class.

The idea is that, maybe, it can be useful in order to use the
configuration class as a container for dry-auto_inject (which expects
`#[]` interface) for user configured dependencies.

Also, it facilitates testing when the configuration class is resolved
from a container, because it allows to be easily interchanged with a
hash. Used in this way (but not if the configuration class dependency is
 hardcoded) it could be a replacement for #19 

The code could be refactored and cleaned a little bit extracting to a
method and using `private_class_method`, but I didn't do it because
I don't know what you would prefer. If the PR is accepted and you prefer
having the code refactored in this way I have no inconvenience in changing
it.

Thanks.

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -17,6 +17,14 @@ module Dry
           end
         end
 
+        klass.__send__(:define_method, '[]') do |name|
+          __send__(name)
+        end
+
+        klass.__send__(:define_method, '[]=') do |name, value|
+          __send__("#{name}=", value)
+        end
+
         klass.new(settings)
       end
 
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -17,14 +17,6 @@ module Dry
           end
         end
 
-        klass.__send__(:define_method, '[]') do |name|
-          __send__(name)
-        end
-
-        klass.__send__(:define_method, '[]=') do |name, value|
-          __send__("#{name}=", value)
-        end
-
         klass.new(settings)
       end
 
@@ -64,6 +56,14 @@ module Dry
         end
       end
       alias to_hash to_h
+
+      def [](name)
+        public_send(name)
+      end
+
+      def []=(name, value)
+        public_send("#{name}=", value)
+      end
     end
   end
 end
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -74,7 +74,7 @@ module Dry
       end
 
       def setting?(name)
-        @config.key?(name)
+        @config.key?(name.to_sym)
       end
     end
   end
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -58,11 +58,23 @@ module Dry
       alias to_hash to_h
 
       def [](name)
+        raise_unknown_setting_error(name) unless setting?(name)
         public_send(name)
       end
 
       def []=(name, value)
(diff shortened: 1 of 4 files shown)


AMHOL:
Thanks for the pr @waiting-for-dev, the idea seems good to me.

Pretty sure these could just be defined as instance methods directly on `Dry::Configurable::Config` like so:

```ruby
def [](name)
  public_send(name)
end

def []=(name, value)
  public_send("#{name}=", value)
end
```

However, with the current implementation we will get a `NoMethodError` when using the hash accessor with an undefined setting name, this should probably be changed to raise an `ArgumentError` or something.

@flash-gordon @fran-worley @solnic @timriley could I get your opinions on this please?

waiting-for-dev:
> Pretty sure these could just be defined as instance methods directly on Dry::Configurable::Config like so:

Haha... Yes, I think I was too obfuscated learning how the library works :)

I pushed the implementation as instance methods. I will wait for the feedback to implement the error handling part (didn't thought about that).

waiting-for-dev:
I just pushed the error handling part.

Now, trying to access an unknown setting through `#[]` or `#[]=` will raise an `ArgumentError` with message `"+#{name}+ is not a setting name"`.

However, there is another aspect to consider. Settings are stored as symbols, so if there is a setting called `:db` and someone tries to access `config['db']` it would raise the error because `'db'` is not a setting name (a `@config` key). If this goes on, what should we do? Should we coerce string keys to symbols or rather make more explicit the error message (something like `"+#{name}+ is not a setting name or it should be accessed as a symbol"`.

AMHOL:
Hmm, good point, I think we should coerce into string values to take advantage of `frozen_string_literal`, if we changed [this line](https://github.com/dry-rb/dry-configurable/blob/master/lib/dry/configurable/config/value.rb#L12) to `name.to_s` and [this line](https://github.com/dry-rb/dry-configurable/pull/22/files#diff-ac529e6c6ac0f3ea46c26f44a7a9d1b0R77) to `@config.key?(name.to_s)` I think that should work.

What are your thoughts on this?

waiting-for-dev:
The other option would be to coerce strings to symbols, just changing [this line](https://github.com/dry-rb/dry-configurable/pull/22/files#diff-ac529e6c6ac0f3ea46c26f44a7a9d1b0R77) to `@config.key?(name.to_sym)`.

I'm not sure which solution wins in the trade-off competition. As fas as I know:

- If we coerce to strings we will benefit from next ruby 3 default frozen strings and from users which have opted-in in 2.3. Users with < 2.3 are out of the game, but still have GC.

- If we coerce to symbols we are always in the frozen side, but users with ruby < 2.2 do not have GC for them.

AMHOL:
Yeah, good point, maybe just go for `to_sym` then, nice one :smile:

waiting-for-dev:
Ok, here it goes!

Good catch about settings set as strings :smile:

AMHOL:
@waiting-for-dev invited you as a collaborator, feel free to push branches for any PRs in future.

Also released as 0.5.0

waiting-for-dev:
Hey, thank you very much @AMHOL , for the quick release and mainly for inviting me as a collaborator. It will be a great pleasure :smile:

AMHOL:
No problem, thanks for the help maintaining the gem :wink:

PR #24, opened 2016-12-28
Add method to reset configuration to default values

waiting-for-dev:
Instead of just removing `@_config`, added method eagerly recreates it
when invoked. This is in order to be 'safe' against inspection through
methods non intended to be public (like directly accessing `@_config`).

Please, notice that line:
```
@_config ||= ::Dry::Configurable::Config.create(_settings) unless _settings.empty?
```
has been changed to:
```
@_config = ::Dry::Configurable::Config.create(_settings) unless _settings.empty?
```
so, memoization has been removed. If I'm not wrong, it is still safe and there is no performance lost because `config` method is guarding with:
```
return @_config if defined?(@_config)
```

@solnic in issue #19 you said:
> We can add it as a test-specific interface that people could include on demand

How would you implement it?

Thanks

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -53,15 +53,6 @@ module Dry
       create_config
     end
 
-    # Resets configuration to default values
-    #
-    # @return [Dry::Configurable::Config]
-    #
-    # @api public
-    def reset_config
-      create_config
-    end
-
     # Return configuration
     #
     # @yield [Dry::Configuration::Config]
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -50,9 +50,16 @@ module Dry
     # @api public
     def config
       return @_config if defined?(@_config)
-      @_config_mutex.synchronize do
-        @_config ||= ::Dry::Configurable::Config.create(_settings) unless _settings.empty?
-      end
+      create_config
+    end
+
+    # Resets configuration to default values
+    #
+    # @return [Dry::Configurable::Config]
+    #
+    # @api public
+    def reset_configuration
+      create_config
     end
 
     # Return configuration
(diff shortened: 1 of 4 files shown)


AMHOL:
@waiting-for-dev thanks for the PR :smile:

You're right about the memoization of `@_config` being unnecessary, seems like it was just an oversight, nice spot.

As for the PR, I feel like `reset_configuration` should be a bang method (`reset_configuration!`), what do you think?

And regarding the test-specific interface, we have something similar in dry-container, you can check out the [PR that introduced it](https://github.com/dry-rb/dry-container/pull/11) if you wanted to go down a similar route.

AMHOL:
Yeah, I normally think of a bang method as an indication of mutation (like in the core language) as opposed to the Rails-ism of a "dangerous" method, although we have a few mutation methods in this gem that don't have the bang suffix - I do feel it would be appropriate in this case, but I'm indifferent really, so take your pick.

`reset_config` (or `reset_config!`) is more consistent, as you said.

timriley:
Feels to me like it'd be best to follow the long-held ruby convention of only offering bang methods if they're a more-"dangerous" alternative to an existing non-bang method of the same name?

AMHOL:
If we don't go with the bang method I need to remember to change [this](https://github.com/dry-rb/dry-container/blob/master/lib/dry/container/stub.rb#L43) (Just a reminder for myself lol)

waiting-for-dev:
Ok, here it goes. Do we need a `dry-interfaces` new friend? :smile:

waiting-for-dev:
Just a note to point to three places where this feature could be useful in `dry-container` tests: [1](https://github.com/dry-rb/dry-container/blob/0929dd26ba66a37329cf31081c44263948833649/spec/support/shared_examples/container.rb#L23), [2](https://github.com/dry-rb/dry-container/blob/0929dd26ba66a37329cf31081c44263948833649/spec/support/shared_examples/container.rb#L62) and [3](https://github.com/dry-rb/dry-container/blob/0929dd26ba66a37329cf31081c44263948833649/spec/support/shared_examples/container.rb#L98)

AMHOL:
Nice one @waiting-for-dev

waiting-for-dev:
Thanks @AMHOL :)

PR #26, opened 2017-02-02
Raise when setting defined on already configured class

AMHOL:
Also removed inheritance of config, this is weird:

```ruby
class Test
  extend Dry::Configurable

  setting :name
end

Test.configure do |config|
  config.name = 'blah'
end

class Test2 < Test; end

Test2.config.name
# => "blah"
```

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -1,6 +1,7 @@
 require 'concurrent'
 require 'dry/configurable/config'
 require 'dry/configurable/config/value'
+require 'dry/configurable/error'
 require 'dry/configurable/version'
 
 # A collection of micro-libraries, each intended to encapsulate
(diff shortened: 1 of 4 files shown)


timriley:
@AMHOL removing the inheritance of the configuration applied on the superclass would break a lot of established conventions in dry-web apps. For example, we use this a lot in dry-view classes, e.g.

```ruby
module MyApp
  class ViewController < Dry::View::Controller
    configure do |config|
      config.layout = "application"
    end
  end

  module Views
    class Home < MyApp::ViewController
      # Inherits "application" layout config from superclass
      configure do |config|
        config.template = "home"
      end
    end
  end
end
```

What's your thinking for removing that behaviour? It's pretty useful for things like this example.

AMHOL:
Ahh OK, I guess I just never thought about it before and when I did it seemed like weird behaviour to me.

AMHOL:
Closing as this was stupid of me :laughing:

PR #27, opened 2017-02-20
Dry::Configurable::TestInterface::reset_config won't reset nested settings

spectator:
Hi,

This PR adds falling test to demonstrate the issue with `Dry::Configurable::TestInterface::rest_config` that won't reset nested settings.

### DIFF
--- a/spec/support/shared_examples/configurable.rb
+++ b/spec/support/shared_examples/configurable.rb
@@ -249,9 +249,13 @@ RSpec.shared_examples 'a configurable class' do
       describe 'reset_config' do
         before do
           klass.setting :dsn, nil
+          klass.setting :pool do
+            setting :size, nil
+          end
 
           klass.configure do |config|
             config.dsn = 'sqlite:memory'
+            config.pool.size = 5
           end
 
           klass.reset_config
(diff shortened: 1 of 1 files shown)


GustavoCaso:
@spectator I have been looking and it looks when using `configure` with nested keys it will not store the default value instead it store the `config` object with the value from the `configure` block.

```ruby
[#<Dry::Configurable::Config::Value:0x007fc6099b8120 @name=:dsn, @value=nil, @processor=#<Proc:0x007fc60b0e2c10@/Users/morty/dev/gems/dry-configurable/lib/dry/configurable/config.rb:5 (lambda)>>, #<Dry::Configurable::Config::Value:0x007fc609a23510 @name=:pool, @value=#<#<Class:0x007fc60aa230f0>:0x007fc60aa21fc0 @config={:size=>5}>, @processor=#<Proc:0x007fc60b0e2c10@/Users/morty/dev/gems/dry-configurable/lib/dry/configurable/config.rb:5 (lambda)>>]
``` 

I do not have much experience with this library, but I think maybe there will need to have some changes regarding the internal structure and how they handle nested attributes.

@AMHOL I'll be happy to give a hand with this one. I was thinking instead of returning the actual proc when calling `config` 

```ruby
#<#<Class:0x007fc60b2f2708>:0x007fc60b2f23e8 @config={:dsn=>"sqlite:memory", :pool=>#<#<Class:0x007fc60aa230f0>:0x007fc60aa21fc0 @config={:size=>5}>}>
``` 

We could return an `OpenStruct` which will allow the user to access with the intended interface:
`klass.config.pool.size`. That wey we could store the actual default value for the nested attributes.

PR #28, opened 2017-02-21
Delegate all NestedConfiguration to its own class to fix reset_config

GustavoCaso:
Regarding the commit https://github.com/dry-rb/dry-configurable/commit/fd8ccfe68aaa5207f0e543b1cc3f92b54d51c5cf where it exposes that `reset_config` was not working for nested configurations.

@AMHOL I have followed what we talked yesterday, and even it since a little by hacky as well, all the logic for nested configurations is encapsulated in its own class.

The PR still in WIP because I wanted to know if you like this approach if so I will add a test for everything that is new and remove the WIP label so we can merge it, if not I will keep looking for an alternative.

@AMHOL thanks for the help.

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -113,9 +113,7 @@ module Dry
 
     # @private
     def _config_for(&block)
-      config_klass = ::Class.new { extend ::Dry::Configurable }
-      config_klass.instance_eval(&block)
-      create_nested_config(config_klass)
+      ::Dry::Configurable::NestedConfig.new(&block)
     end
 
     # @private
@@ -134,11 +132,6 @@ module Dry
     # @private
     def nested_configs
       _settings.select { |setting| setting.value.kind_of?(::Dry::Configurable::NestedConfig) }.map(&:value)
-    end
-
-    # @private
-    def create_nested_config(klass)
-      ::Dry::Configurable::NestedConfig.new(klass)
     end
   end
 end
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -1,5 +1,6 @@
 require 'concurrent'
 require 'dry/configurable/config'
+require 'dry/configurable/nested_config'
 require 'dry/configurable/config/value'
 require 'dry/configurable/version'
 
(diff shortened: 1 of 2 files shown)


lib/dry/configurable/nested_config.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+      end
+
+      private
+
+      def config
+        @klass.config
+      end
+
+      def method_missing(method, *args, &block)
```
AMHOL: Never seen this defined as a private method before, any implications?

lib/dry/configurable.rb
```diff
@@ -114,14 +115,30 @@ def _settings
     def _config_for(&block)
       config_klass = ::Class.new { extend ::Dry::Configurable }
       config_klass.instance_eval(&block)
-      config_klass.config
+      create_nested_config(config_klass)
```
AMHOL: Perhaps the above two lines can be moved to the new `NestedConfig` class too, an this could just become:

```ruby
::Dry::Configurable::NestedConfig.new(&block)
```

lib/dry/configurable/nested_config.rb
GustavoCaso: If I'm honest no idea, but I can move outside of `private`

AMHOL: It's fine as-is, I guess method missing will only be called privately anyway, just never thought of it before :smile:

lib/dry/configurable.rb
GustavoCaso: But we still have to pass the `config_klass`instance to `::Dry::Configurable::NestedConfig`.

AMHOL: Could just update `::Dry::Configurable::NestedConfig.new` to:

```ruby
module Dry
  module Configurable
    # @private
    class NestedConfig
      def initialize(&block)
        klass = ::Class.new { extend ::Dry::Configurable }
        klass.instance_eval(&block)
        @klass = klass
      end
      # ...
```

Then it fully encapsulates the creation of nested config

GustavoCaso: Awesome thats great I will update the code, and the missing test so we can move forward

GustavoCaso:
@AMHOL, since is a `NestedConfig` doesn't have much to test in my opinion, and there is the complete integration test, for the `configurable` Module we could ship this.

What do you think?

AMHOL:
@GustavoCaso seems fair to me, thanks for implementing this :+1:

AMHOL:
@GustavoCaso this is now released under `0.6.1`

Issue #30, opened 2017-02-21
Raise when a setting is defined after config is created


AMHOL:
Currently if you try to define a setting after config has been created it will just be dropped silently, we should raise if `setting` is called after `create_config`

GustavoCaso:
@AMHOL I think we can close this one.

AMHOL:
:+1: Thanks

PR #31, opened 2017-02-22
Raise exception for alreday created configs

GustavoCaso:
#30 

@AMHOL raise an exception when try to set a new configuration after `create_config` has been called.

I saw in the closed pull request there was one with a similar description, so I continue with the error definition style you had in mind, hope you like it 😄.

Also, I had to modify some tests in order to make it pass regarding when inheriting.

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -139,7 +139,7 @@ module Dry
     # @private
     def raise_already_defined_config(key)
       raise AlreadyDefinedConfig,
-        "Cannot add setting `#{key}`, #{self} is already configured"
+        "Cannot add setting +#{key}+, #{self} is already configured"
     end
   end
 end
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -1,5 +1,6 @@
 require 'concurrent'
 require 'dry/configurable/config'
+require 'dry/configurable/error'
 require 'dry/configurable/nested_config'
 require 'dry/configurable/config/value'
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples/configurable.rb
```diff
@@ -224,20 +224,20 @@
...
 
-        context 'when the inherited settings are modified' do
+        context 'when the inherited settings are modified ' do
           before do
             klass.setting :dsn
             klass.configure do |config|
               config.dsn = 'jdbc:sqlite:memory'
             end
-
-            subclass.setting :db
```
AMHOL: Does this test not pass is this is executed before `klass.configure` is called above?

lib/dry/configurable.rb
```diff
@@ -133,5 +135,11 @@ def create_config_for_nested_configurations
     def nested_configs
       _settings.select { |setting| setting.value.kind_of?(::Dry::Configurable::NestedConfig) }.map(&:value)
     end
+
+    # @private
+    def raise_already_defined_config(key)
+      raise AlreadyDefinedConfig,
+        "Cannot add setting `#{key}`, #{self} is already configured"
```
AMHOL: Could we update this to be more in line with https://github.com/dry-rb/dry-configurable/blob/master/lib/dry/configurable/config.rb#L73 please? Or vice versa, not bothered which.

i.e. `Cannot add setting +#{key}+, #{self} is already configured`

GustavoCaso: Sure

Issue #25, opened 2017-01-20
Reader pre-processor


waiting-for-dev:
We already have a writer pre-processor:

```ruby
setting(:file) { |value| Pathname(value) }
```

But now, I'm in a use case where a reader pre-processor could also be useful.  In my case I would like to throw an error if an important setting is not set. I think it would be nice to do it in the configuration layer, because it is its outermost boundary.

I see that having a setting with both writer and reader pre-processors could look a bit clumsy:

```ruby
setting(:file,
        reader: ->(value) { value || raise('+file+ setting must be defined') }
       ) { |value| Pathname(value) }
```

If we are not in a "don't break backwards compatibility" phase, we could instead do:

```ruby
setting(:file,
        reader: ->(value) { value || raise('+file+ setting must be defined') },
        writer: ->(value) { Pathname(value) })
```

Or, in order not to break compatibility, we could make a single block default to a writer pre-processor.

What do you think about it?

AMHOL:
@waiting-for-dev I like this idea, going to add 1.0 milestone and then we can break what we like lol

FWIW for 1.0 I also plan to address #8 and change the default value to a keyword argument (`default`)

waiting-for-dev:
Sounds good ;)

So, do you also think having both `reader` & `writer` options would be the best syntax? I think it is the most explicit and clear.

AMHOL:
Yeah I think so too, then it could just default to identity function

GustavoCaso:
@waiting-for-dev are you currently working on it? If not I would like to have a look see if I can implement it.

waiting-for-dev:
@GustavoCaso feel free to take it

GustavoCaso:
Awesome @waiting-for-dev

AMHOL:
BTW I think that we should use `:setter` and `:getter` rather than `:reader` and `:writer` so that we can leave `:reader` and `:writer` open for #8.

waiting-for-dev:
Hmm... I don't like `setter` and `getter` nomenclature... it's not that important, but I think is not idiomatic in Ruby. More importantly, it would be quite confusing to remember which pair does what...

If #8 is going to be implemented (I'm not sure it is a good idea), maybe we could call that pair `forward_reader` and `forward_writter` ?

AMHOL:
Perhaps we could just go with something like:

```ruby
setting :path,
        default: Dir.pwd,
        reader: ->(value) { value || raise('+file+ setting must be defined') },
[...]

Issue #8, opened 2016-08-15
Reader option


solnic:
I've found myself doing that a lot:

``` ruby
class Foo
  extend Dry::Configurable

  setting :something

  def self.something
    config.something
  end
end
```

WDYT about this:

``` ruby
class Foo
  extend Dry::Configurable

  setting :something, reader: true
end
```

artofhuman:
What behavior for reader should be for this example?
 ``` ruby
class Foo 
  extend Dry::Configurable
  
  setting :database, reader: true do
    setting :dsn, 'sqlite:memory'
  end
end

AMHOL:
I'm not sure it would be necessary on nested settings as a reader on the root setting would make the child settings available as i.e. `database.dsn`

AMHOL:
Not sure about this as it introduces different behaviour based on the level of nesting, plus you could just use Forwardable and delegators to do the same

solnic:
@AMHOL wdym?

timriley:
@AMHOL's probably talking about what happens with something like this, for example?

```ruby
setting :foo do
  setting :bar, reader: true
end
```

solnic:
Nesting shouldn't be supported as it makes no sense :)

AMHOL:
@solnic I'm not sure the option should be supported though, the way the library works it can't be implemented on the root level without also being implemented at the nested level, and I'm not sure whether a library for configuration should support an option to expand the interface of an object. With `Forwardable` I meant something like:

```ruby
class App
  extend Dry::Configurable

  class << self
    extend Forwardable
    def_delegator :config, :database
  end

  # Pass a block for nested configuration (works to any depth)
  setting :database do
    # Can pass a default value
    setting :dsn, 'sqlite:memory'
  end
  # Defaults to nil if no default value is given
  setting :adapter
end

App.database
=> #<struct dsn="sqlite:memory">
```

GustavoCaso:
@AMHOL I think we can close this one

PR #33, opened 2017-02-24
Add reader option to create accessor method at the class level

GustavoCaso:
#8 #25 

Since the change for accepting `Dry::Types` looks a little big, I thought about dividing the task in different pull requests.
With this experiment, I used the `method_missing` to create the reader method, at first I use `define_method` but it was more verbose.

One thing I found with this solution, is that for `setting` with a `block` you have to pass the default value even if it is `nil`
```ruby
klass.setting(:dsn, 'memory', reader: true) { |dsn| "sqlite:#{dsn}" }
klass.dsn # => 'sqlite:memory'

klass.setting(:dsn, nil, reader: true) { |dsn| "sqlite:#{dsn}" }
klass.dsn # => 'sqlite:'
```
Because with the definition of `setting`
```ruby
def setting(key, value = ::Dry::Configurable::Config::Value::NONE, options = {}, &block)
```
the second argument always is going to be the value.
We could use `keyword` arguments like:
```ruby
def setting(key, value: ::Dry::Configurable::Config::Value::NONE, reader: false, &block)

klass.setting(:dsn, reader: true) {|dsn| "sqlite:#{dsn}" }
klass.dsn #=> nil

klass.setting(:dsn, value: 'memory', reader: true) { |dsn| "sqlite:#{dsn}" }
klass.dsn # => 'sqlite:memory'
```

But this will break the public api, I want to know your opinion first.
P.D I left a skipping test, that shows the behavior described above.
@solnic @AMHOL

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ class App
   # Defaults to nil if no default value is given
   setting :adapter
   # Passing the reader option as true will create reader method for the class
-  setting :pool, 5 , reader: true
+  setting :pool, 5 ,reader: true
   # Passing the reader attributes works with nested configuration
-  setting :uploader, nil, reader: true do
+  setting :uploader ,reader: true do
     setting :bucket, 'dev'
   end
 end
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ class App
   end
   # Defaults to nil if no default value is given
   setting :adapter
+  # Passing the reader option as true will create reader method for the class
+  setting :pool, 5 , reader: true
+  # Passing the reader attributes works with nested configuration
+  setting :uploader, nil, reader: true do
+    setting :bucket, 'dev'
+  end
 end
 
 App.configure do |config|
@@ -34,6 +40,8 @@ end
 App.config.database.dsn
 # => 'jdbc:sqlite:memory'
 App.config.adapter # => nil
+App.pool # => 5
+App.uploader.bucket # => 'dev'
 ```
 
 ## Links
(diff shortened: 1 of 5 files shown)


solnic:
we can change the signature to be `def setting(key, *args, **opts, &block)` and this won't break existing API

GustavoCaso:
Awesome @solnic thanks

GustavoCaso:
Looks like for values like hashes it does not work

```ruby 
klass.setting(:db_config, { user: 'root', password: '' }
def setting(key, *args, **opts, &block)
args # => []
opts # => {:user=>"root", :password=>""}
```

solnic:
I thought for hashes we prefer a nested block syntax so:

``` ruby
setting :db_config do
  setting :user, 'root'
  setting :password, ''
end

# then 
config.db_config.user
config.db_config.password
```

FWIW I prefer this syntax over using hashes as setting values

GustavoCaso:
@solnic @AMHOL I have updated the code to accept both arguments and options. Please fell free to add any comment.

Thanks.

solnic:
LGTM

README.md
```diff
@@ -25,6 +25,12 @@ class App
   end
   # Defaults to nil if no default value is given
   setting :adapter
+  # Passing the reader option as true will create reader method for the class
+  setting :pool, 5 ,reader: true
```
AMHOL: Could we fix the formatting here and below please

AMHOL: In fact, it's only minor so I'll just push a commit to master

AMHOL:
Nice one @GustavoCaso

GustavoCaso:
Thanks @AMHOL was great working on this

PR #34, opened 2017-03-05
[WIP] Added Support for `dry-types`

GustavoCaso:
#25 

@AMHOL as we talk about it I have created this PR so we can discuss design decisions, to add this feature to the lib.

Once we are happy with this we can start working on the new dsl for `dry-types`

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -105,11 +105,15 @@ module Dry
         end
       end
 
+      value, preprocessor = extract_preprocessor(value)
+
       _settings << ::Dry::Configurable::Config::Value.new(
         key,
         !value.nil? ? value : ::Dry::Configurable::Config::Value::NONE,
-        processor || ::Dry::Configurable::Config::DEFAULT_PROCESSOR
+        processor || ::Dry::Configurable::Config::DEFAULT_PROCESSOR,
+        preprocessor
       )
+
       store_reader_options(key, options) if options.any?
     end
 
@@ -132,6 +136,10 @@ module Dry
     end
 
     private
+
+    def extract_preprocessor(value)
+      value.respond_to?(:call) ? [nil, value] : [value, nil]
+    end
 
     # @private
     def _config_for(&block)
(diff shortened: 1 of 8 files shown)


lib/dry/configurable.rb
```diff
@@ -94,11 +94,15 @@ def setting(key, *args, &block)
...
       end
 
+      value, preprocessor = extract_preprocessor(value)
+
       _settings << ::Dry::Configurable::Config::Value.new(
         key,
         value || ::Dry::Configurable::Config::Value::NONE,
-        processor || ::Dry::Configurable::Config::DEFAULT_PROCESSOR
+        processor || ::Dry::Configurable::Config::DEFAULT_PROCESSOR,
+        preprocessor
```
GustavoCaso: I don't know if I really like the idea having a `preprocessor`, but for the example it makes is use. So please feel free to comment as much as need it

lib/dry/configurable.rb
```diff
@@ -122,6 +126,10 @@ def _reader_attributes
 
     private
 
+    def extract_preprocessor(value)
+      value.respond_to?(:call) ? [nil, value] : [value, nil]
```
GustavoCaso: I know it doesn't explain much but I will improve the code probably in the `ArgumentParser` class to handle all this logic

AMHOL:
Personally I think we might be better off making dry-types a mandatory dependency, that way it can handle any processing and setting the default value, i.e.

```ruby
class Something
  extend ::Dry::Configurable

  setting :dsn, Types::Strict::String.default('sqlite::memory')
end
```

solnic:
👍 for making dry-types a dependency, it can encapsulate a lot of functionality and it's already a dependency in other dry-rb gems and rom-rb anyway. It would also be another use case and should help in making dry-types more robust and feature-complete.

solnic:
@GustavoCaso could you rebase this / resolve conflicts?

GustavoCaso:
@solnic @AMHOL adding `dry-types` as a dependency maybe is not that trivial since `dry-configurable` is already a dependency in `dry-types`, so what I did is make it work an interface similar to `dry-types` `#call`.

WDYT?

solnic:
We should remove dry-configurable dep from dry-types, it's an overkill. We only use it to set `:namespace`, which we could simply do via `Dry::Core::ClassAttributes` instead of configurable DSL.

GustavoCaso:
We can close this one. The use of dry-types was added with 1caa28e3b405db806aaf1dc09748e1213ec9bd08

PR #35, opened 2017-03-18
allow a boolean false as default value

yuszuv:
I think it makes sense to allow a boolean value like `false` as a default value, so I implemented it :)


GustavoCaso:
Looking at the code it totally makes sense. Let see what @AMHOL have to say about this. Maybe is like this by design

AMHOL:
Nope, not by design, thanks :smile:

PR #38, opened 2017-03-20
Disallow modification on frozen config

qcam:
I think we shall disallow modifying configs when it has been frozen or finalized

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 ## Unreleased
 
+## Added
+
+* Introduce `Configurable.finalize!` which freezes config and its dependencies ([qcam](https://github.com/qcam))
+
 ## Fixed
 
 * Allow for boolean false as default setting value ([yuszuv](https://github.com/yuszuv))
 * Convert nested configs to nested hashes with `Config#to_h` ([saverio-kantox](https://github.com/saverio-kantox))
+* Disallow modification on frozen config ([qcam](https://github.com/qcam))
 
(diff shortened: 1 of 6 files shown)


AMHOL:
Thanks @qcam, looks good, could you update `CHANGELOG.md` please

qcam:
Thanks @AMHOL, CHANGELOG updated and commits squashed.

AMHOL:
Thanks @qcam 

Sorry I didn't think before but it might be better to define a custom error for this in https://github.com/dry-rb/dry-configurable/blob/master/lib/dry/configurable/error.rb, what do you think?

qcam:
Custom error sounds cool to me! 👍 

Do you think we want a custom `freeze` method like `finalize!` in dry-container?

AMHOL:
Yeah I think that would be good, could just call `freeze` and `@config.freeze`

CHANGELOG.md
```diff
@@ -4,5 +4,7 @@
 
 * Allow for boolean false as default setting value ([yuszuv](https://github.com/yuszuv))
 * Convert nested configs to nested hashes with `Config#to_h` ([saverio-kantox](https://github.com/saverio-kantox))
+* Introduce `Configurable.finalize!` which freezes config and its dependencies ([qcam](https://github.com/qcam))
```
AMHOL: Can you add this under a `## Added` title please

P.S. Sorry to be a pain and keep asking for changes

qcam: @AMHOL no worries, changelog updated as requested

Issue #43, opened 2017-11-17
Nested configs aren't finalized


dikond:
Environment:
```ruby
require 'dry-configurable'

class App
  extend Dry::Configurable

  setting :nested do
    setting :foo, 'bar'
  end

  setting :deeply_nested do
    setting :foo do
      setting :bar, 'baz'
    end
  end
end
```

Expected behaviour:
```ruby
App.config.finalize! # or App.finalize!

App.config.nested.foo = 'oops'
# => Dry::Configurable::FrozenConfig: Cannot modify frozen config

App.config.deeply_nested.foo.bar = 'oops'
# => Dry::Configurable::FrozenConfig: Cannot modify frozen config
```

Right now, however, no errors are raised and we can easily modify the config.

GustavoCaso:
With the latest changes in the library, this no longer can occur.

PR #44, opened 2017-11-17
Fix finalization bug

dikond:
Fix for issue #43


GustavoCaso:
@dikond with the latest changes in master this bug is no longer an issue, the internals of the library has changed to use `dry-struct` to create the config objects, so you can no longer modify the config objects. Thank you so much for opening the PR ❤️

PR #48, opened 2018-02-13
Dry struct as config object

GustavoCaso:
This PR adds `dry-struct` to use it as the config object.

This PR remove the old code and simplify the design of the library by using `dry-struct` and `dry-types` extensively.

Features:
- [x] Use `dry-struct`
- [x] Allow nested configuration
- [x] Use of defaults values from `dry-types`
- [x] Processor can be achieved with `dry-types` constructor methods, provided tests 
- [x] Add `reader` option but could be achieved with `dry-types` meta attributes, provided tests 
- [x] Freeze decided by user
- [ ] Works when inherited
- [ ] Split test in small files for better readability
- [ ] Documentation

Questions:
- Using reader option from the meta attributes in nested settings:
   What do we prefer to set the value on the parent block or in the inner block
   ```ruby
   Class.new do
      extend Dry::Configurable

      setting :database, Types::Hash.meta(reader: true) do
        setting :url, Types::String.meta
      end
   end
   ```
    or 
    ```ruby
     Class.new do
        extend Dry::Configurable

        setting :database do
          setting :url, Types::String.meta(reader: true)
        end
    end
    ```

In the current implementation, I went with the second approach the issue I have to loop over all the struct schema to get all the keys and see if they have the reader attribute set.

### DIFF
--- a/lib/dry/configurable.rb
+++ b/lib/dry/configurable.rb
@@ -1,4 +1,5 @@
 require 'dry-struct'
+require 'dry/configurable/config'
 require 'dry/configurable/null_config'
 
 module Dry
@@ -27,20 +28,6 @@ module Dry
     NotConfiguredError  = Class.new(StandardError)
     AlreadyDefinedConfigError = Class.new(StandardError)
     FrozenConfigError = Class.new(StandardError)
-
-    class Config < Dry::Struct
-      class << self
-        private :attribute
-
-        def setting(name, type = nil, &block)
-          if block
-            attribute(name, Class.new(Config), &block)
-          else
-            attribute(name, type)
-          end
-        end
-      end
-    end
 
     def self.extended(klass)
       klass.class_eval do
--- a/lib/dry/configurable/null_config.rb
+++ b/lib/dry/configurable/null_config.rb
@@ -2,7 +2,7 @@ module Dry
   module Configurable
 
     # @private
-    class NullConfig < Dry::Struct
+    class NullConfig < Config
       transform_types do |type|
         if type.is_a?(Class) && type <= Dry::Struct || type.default?
           type
@@ -14,14 +14,8 @@ module Dry
       end
 
       class << self
-        private :attribute
-
         def setting(name, type = nil, &block)
-          if block
-            attribute(name, Class.new(NullConfig), &block)
-          else
-            attribute(name, type)
-          end
+          super
 
           define_method("#{name}=") do |value|
             @attributes = @attributes.merge(name => value)
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -1,0 +1,17 @@
+module Dry
+  module Configurable
+    class Config < Dry::Struct
+      class << self
+        private :attribute
+
+        def setting(name, type = nil, &block)
+          if block
+            attribute(name, Class.new(self.superclass), &block)
+          else
+            attribute(name, type)
+          end
+        end
+      end
+    end
+  end
+end


AMHOL:
Been discussing this with @GustavoCaso in PMs on Gitter, I think we could do something roughly along the lines of the following, allowing us to maintain the same interface in the `configure` block:

```ruby
class Config < Dry::Struct
  class << self
    private :attribute

    def setting(name, type = nil, &block)
      if block
        attribute(name, Class.new(Config), &block)
      else
        attribute(name, type)
      end

      define_method("#{name}=") { |value| attributes[name] = value }
    end

    def configure
      yield(config)
      finalize!
    end

    def config
      @config ||= NullConfig.from(self)
    end

    def finalize!
      @config = config.finalize!
    end
  end
end

class NullConfig < SimpleDelegator
  class << self
    def from(struct_class)
     config_class = Class.new(struct_class)
     null_config_class = nullify(config_class)
     new(null_config_class.new, struct_class: struct_class)
    end

    private

    def nullify(config_class)
      config_class.tap do |config|
        config.transform_types do |type|
          if type.respond_to?(:transform_types)
            nullify(type)
          else
            type.meta(omittable: true)
          end
        end
      end
    end
  end

  attr_reader :struct_class

  def initialize(null_config, struct_class: struct_class)
    super(null_config)
    @struct_class = struct_class
  end

  def finalize!
    struct_class.new(__getobj__)
  end
end
```

WDYT @solnic @timriley @flash-gordon

GustavoCaso:
With the latest change introduce in commit a171eac
We are able to keep the same DSL as we have before.

```ruby
Class.new do
  extend Dry::Configurable

  setting :database do
    setting :url, Test::Types::String.meta(reader: true)
  end
end

klass.configure do |config|
  config.database.url = 'localhost'
end
```

lib/dry/configurable.rb
```diff
@@ -51,9 +51,13 @@ def setting(name, type = nil, &block)
       store_reader_key(name) if reader_options?
     end
 
-    def configure(&block)
-      settings_values = ProxySettings.new(struct_class, &block).schema
-      @config = struct_class.new(settings_values)
+    def configure
+      yield(null_config)
```
AMHOL: This needs to yield the config struct on subsequent calls, otherwise the config will always be blank on calls to `configure`

lib/dry/configurable/null_config.rb
```diff
@@ -0,0 +1,40 @@
...
+            hash[key] = value
+          end
+        end
+      end
+
+      EQUAL_END = /=$/
+
+      def method_missing(method_name, *args, &block)
+        to_return = self
+        if method_name.to_s.match(EQUAL_END)
```
splattael: Using `rindex` + `string[0, index]` might be faster:

### Ref

https://github.com/neopoly/nested_config/blob/7b8c1f7ccf9ad14ec02cb9fdefbf7fa86a925d14/lib/nested_config.rb#L79,L86

### Bench

```ruby
require "benchmark/ips"

EQUAL_END = /=$/

def mm(method_name)
  key = method_name.to_s

  if method_name.to_s.match(EQUAL_END)
    key = method_name.to_s.gsub(EQUAL_END, '').to_sym
  else
    # noop
  end
end

EQUAL = "=".freeze

def mm2(method_name)
  key = method_name.to_s
  index = key.rindex(EQUAL)
  if index
    key = key[0, index].to_sym
  else
    # noop
  end
end

p mm(:method)
p mm(:method=)

p mm2(:method)
p mm2(:method=)

Benchmark.ips do |x|
  x.report "mm getter" do
    mm(:method)
  end

  x.report "mm setter" do
    mm(:method=)
  end

  x.report "mm2 getter" do
    mm2(:method)
  end

  x.report "mm2 setter" do
    mm2(:method=)
  end

  x.compare!
end
```

### Results

```
nil
:method
nil
:method
Warming up --------------------------------------
           mm getter    83.252k i/100ms
           mm setter    21.563k i/100ms
          mm2 getter   140.865k i/100ms
          mm2 setter    96.049k i/100ms
           mm getter      1.455M (± 2.0%) i/s -      7.326M in   5.037411s
           mm setter    267.059k (± 2.5%) i/s -      1.337M in   5.009369s
          mm2 getter      2.860M (± 3.5%) i/s -     14.368M in   5.029904s
          mm2 setter      1.551M (± 3.2%) i/s -      7.780M in   5.020025s

Comparison:
          mm2 getter:  2860183.6 i/s
          mm2 setter:  1551423.7 i/s - 1.84x  slower
           mm getter:  1454968.8 i/s - 1.97x  slower
           mm setter:   267059.2 i/s - 10.71x  slower
```

lib/dry/configurable.rb
GustavoCaso: @AMHOL I think is not necessary since `null_config` will always return the same `NullConfig` object, so no matter how many times you call `configure`. I will add more tests to it so to prove it.

solnic: I thought with improved dry-types we won't have to do much here, to set defaults. @flash-gordon ?

GustavoCaso: @solnic I think defaults on `dry-types` works as expected. Is a matter of what we want `dry-configurable` works. So what I try with this is approach is for cases where the user specifies default for all the key, so in that case, there is no need to call `configure` he can directly call `finalize!` and he will get his config object.

with a simple configuration, there is no problem
```ruby
class User < Dry::Struct
  attribute :hello, Types::String.default('world')
end

User.new
=> #<User hello="world">
```

the problem comes with a nested configuration 
```ruby
class NestedUser < Dry::Struct
  attribute :nested do
    attribute :name, Types::String.default('world')
  end
end

Nesteduser.new
=> Dry::Struct::Error: [NestedUser.new] :nested is missing in Hash input

NestedUser.new({nested: {}})
=> #<NestedUser nested=#<NestedUser::Nested name="world">>
```

So for simplifying the experience to the user if all the nested keys have defaults and he/she do not want to `configure` he/she should be able to just call `finalize!` and get a config object.

@solnic , @flash-gordon , @timriley , @AMHOL WDYT?

solnic: FWIW, I've never ever used nested syntax O_o

lib/dry/configurable/null_config.rb
flash-gordon: I believe `end_with?('=')` + `.chop` will be even faster :)

flash-gordon:
Can't we use default values for this somehow? Although, this means we'd have to transform each non-default type with `.optional.default(nil)`, except for nested structs ofc.

flash-gordon:
"for this" — I mean as a replacement for nullconfig

mensfeld:
What about lambdas? I know that this is not the primary goal of this library but for me it is a critical feature that we use a lot. For example here: https://github.com/karafka/karafka/blob/master/lib/karafka/setup/config.rb#L27

lib/dry/configurable/null_config.rb
splattael: @flash-gordon I like your approach :+1: which not also looks nicer but is tiny bit faster as well :rocket: 

## Bench

```ruby
require "benchmark/ips"

EQUAL = "=".freeze

def mm2(method_name)
  key = method_name.to_s
  index = key.rindex(EQUAL)
  if index
    key = key[0, index].to_sym
  else
    # noop
  end
end

def mm3(method_name)
  key = method_name.to_s
  if key.end_with?(EQUAL)
    key = key.chop.to_sym
  else
    # noop
  end
end

p mm2(:method)
p mm2(:method=)

p mm3(:method)
p mm3(:method=)

Benchmark.ips do |x|
  x.report "mm2 getter" do
    mm2(:method)
  end

  x.report "mm2 setter" do
    mm2(:method=)
  end

  x.report "mm3 getter" do
    mm3(:method)
  end

  x.report "mm3 setter" do
    mm3(:method=)
  end

  x.compare!
end
```

## Results

```
nil
:method
nil
:method
Warming up --------------------------------------
          mm2 getter   213.025k i/100ms
          mm2 setter   151.135k i/100ms
          mm3 getter   227.005k i/100ms
          mm3 setter   169.490k i/100ms
          mm2 getter      3.696M (± 7.0%) i/s -     18.533M in   5.045861s
          mm2 setter      2.146M (± 6.7%) i/s -     10.731M in   5.027797s
          mm3 getter      4.120M (± 8.6%) i/s -     20.430M in   5.010009s
          mm3 setter      2.401M (±11.0%) i/s -     11.864M in   5.021200s

Comparison:
          mm3 getter:  4119677.6 i/s
          mm2 getter:  3695816.6 i/s - same-ish: difference falls within error
          mm3 setter:  2401452.5 i/s - 1.72x  slower
          mm2 setter:  2145832.0 i/s - 1.92x  slower
```

timriley:
> What about lambdas? I know that this is not the primary goal of this library but for me it is a critical feature that we use a lot. For example here: https://github.com/karafka/karafka/blob/master/lib/karafka/setup/config.rb#L27
> 
> Same with classes / modules as values.
> What we basically do is that after everything is defined and optionally reconfigured by the user, we pipe it to dry-validations for checks. Without possibility to use proc and classes as values, it gets really problematic for us to use it :(

Can we make sure we've addressed @mensfeld's concern here? @GustavoCaso, perhaps you could provide an example of how these will remain supported?

timriley:
Can I ask another (very late) question about this PR? Was there a particular reason we're relying on a type object's `meta` for accessing the `:reader` option for a setting?

I just feel like this:

```ruby
setting :my_thing, Types::Strict::String.meta(reader: true)
```

Is not as nice as this, where the setting-specific options are provided separately to the type object:

```ruby
setting :my_thing, Types::Strict::String, reader: true
```

Is something like this possible in this new arrangement? Internally, we might decide to go ahead and store these setting-related options on the type object's `meta` as the way to pass them around, but I feel we offer a worse experience if we're foisting that responsibility onto the end user.

PR #49, opened 2018-02-28
Use dry struct for the config object

GustavoCaso:
@solnic @flash-gordon @timriley @AMHOL 

Here is the 3 attempt 🤣, hope this PR is more clear and we can discuss any issue or things to improve before merging it.

I decided not to delete the unused files just skip the test that way is less noise in the PR, also I decided to update the file `spec/support/shared_examples/configurable.rb` to work with the current `API` instead of rewritten them in a new file, just hope is clear enough to see the changes. I have left some test in the shared example as skip so we can discuss how we want it work

Implementations details:
- `Config` is `dry-struct` that will hold all the config information
- `NullConfig` is same as `Config` but works as `NullObject` thanks to `transform_type` also set setter methods for each attribute that way we can pass it to the `configure` method and keep holding the same `API`
```ruby
config_klass.configure do |config|
  config.database = 'test'
end
```

There are a couple of things that couldn't make it work with the current implementation: 
- Having a setting defined as `Types::Hash` with default values will not work because `dry-struct` will not successfully build the `default_attributes` but as @flash-gordon mentioned to me in private messages doesn't make much sense to store attributes as Hash in `dry-struct`

- By default, if not set values to the settings  it will return `nil` with the new `API` it will return the default value or an error if no valid struct could be built, this is a test example that I have skipped that show the old behavior 
https://github.com/dry-rb/dry-configurable/pull/49/files#diff-4893c57cc8a61b99645d27a46495622eR126

A doubt I had:
- Not sure if this the expected way we want to define `reader` option for nested configurations 
https://github.com/dry-rb/dry-configurable/pull/49/commits/08073ea41ac1fc793d58ad9d3d2bca1787b8a343#diff-4893c57cc8a61b99645d27a46495622eR95

Hope we can discuss it here. I really want to hear your opinions on this one


lib/dry/configurable/config.rb
```diff
@@ -1,91 +1,36 @@
...
-            @config[setting.name] = nil
+        def setting(name, type_or_value = nil, &block)
+          if block
+            attribute(name, Class.new(self.superclass), &block)
           else
-            public_send("#{setting.name}=", setting.value)
+            type = if type?(type_or_value)
+              type_or_value
+            else
+              warn deprecation_warning
```
flash-gordon: pls use `dry/core/deprecations`

PR #52, opened 2018-07-12
Remove execution from `version.rb` file

itsNikolay:


GustavoCaso:
Thank you so much. Nice catch

PR #53, opened 2018-08-01
Depend on `dry-struct` gem

alsemyonov:
Fixes `LoadError`.
See https://github.com/dry-rb/dry-configurable/commit/1caa28e3b405db806aaf1dc09748e1213ec9bd08#r29914373


alsemyonov:
No, we don’t need to have `dry-container` in `dry-configurable.gemspec`, since it is not explicitly used anywhere in `dry-configurable` sources: it is a dependency of `dry-types`, so it is covered by their gemspec file, which in turn is a dependency of `dry-struct`.

GustavoCaso:
Thank you so much @alsemyonov

PR #55, opened 2018-12-19
[WIP] Add conditions

### LINKED ISSUE #54: Allow to set configuration based on conditions
Example:

```ruby
class App
  extend Dry::Configurable

  setting :logger
  setting :production, if: -> { ENV['RACK_ENV'] == 'production' } do
	end
end

App.configure do |config|
  config.logger = { level: :debug }

  config.production do
    config.logger = { level: :info }
  end
end
```

GustavoCaso:
Fix #54 

@timriley this is the POC A I mentioned early.

Of course, this is a WIP, I would have to clean the tests and maybe some other stuff will come in the future, but I wanted to have something we can discuss.

WDYT?

I have one question doe:

* In the test I check for an `ENV` variable but that might not be the case all the time, maybe some user needs to check some other object or another value, that might not be accessible to the `config` at the point of finalizing.  
I do not know if your suggestion of `finalize!(environment: EnvironmentObject.new) ...` was on those lines?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ script:
 after_success:
   - '[ -d coverage ] && bundle exec codeclimate-test-reporter'
 rvm:
-  - 2.2
   - 2.3.1
   - jruby-9.1.5.0
   - ruby-head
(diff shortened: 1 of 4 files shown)


solnic:
Closing due to lack of activity. Furthermore, I'm quite sure we need a major clean up or even a rewrite before adding any new features.

PR #56, opened 2019-01-08
[WIP] Blockless config

timriley:
This is a first step exploring the removal of the `configure` block-based API.


flash-gordon:
@timriley I rebased and force-pushed this branch

flash-gordon:
@timriley specs are passing, what else do you want to explore? 😄

timriley:
@flash-gordon TBH, not a great deal, I think! I think we just need to add an additional spec to cover the old-school `configure do |config|` style, which @AMHOL suggested would be best to keep in.

solnic:
Can we also add an interface to finalize the config? It should freeze all configured values, IMO.

PR #58, opened 2019-01-10
Correct homepage in the gemspec

amatsuda:
It seems that dryrb organization has been renamed to dry-rb at some point in the history, but the gemspec still points to the old URL.


timriley:
Thanks @amatsuda :)

PR #59, opened 2019-01-12
Revert conversion to dry-struct

timriley:
Using dry-struct was neat in some ways, in that a config object is effectively a typed struct, but it introduced unsatisfiable gem dependency cycles, e.g. that dry-struct required dry-types, and dry-types required dry-container, and dry-container was already using dry-configurable.

Secondly, with the dry-struct dependency bringing in those extra transitive depdencies, dry-configurable became less lightweight, and less palatable for others to use, especially given it has such a narrow, low-level focus.

We'll reintroduce type support via another approach which doesn't have these downsides.

This reverts the following commits:

- 1caa28e3b405db806aaf1dc09748e1213ec9bd08
- abefc03b945fb39349461b46b9b3a7aefc77a2ad
- 208e6a426fdf344663265c02e00f052433d36ecf
- 600d60703d27babc90e8051af0ad4538d188eb6b
- 2ecae98d8932b8d215836817a39a5c7b7db52525


.travis.yml
```diff
@@ -8,8 +8,11 @@ script:
 after_success:
   - '[ -d coverage ] && bundle exec codeclimate-test-reporter'
 rvm:
+  - 2.0
+  - 2.1
   - 2.2
   - 2.3.1
+  - rbx-3
```
solnic: How come this is part of this PR? :/

timriley: @solnic looks like rbx-3 was _removed_ as part of https://github.com/dry-rb/dry-configurable/pull/49. Since I'm reverting the _merge commits_ for the recent PRs, it gets added back here.

It's kind of a bother this was changed within the PR, but rather than fiddling around and reverting every individual commit within the PRs (except this one), I might just go back and fix up the .travis.yml afterwards, in a separate commit (it needs new rubies anyway).

solnic: d'ooh of course, thanks!

flash-gordon:
I updated CI config, let's not to hold this anymore

PR #60, opened 2019-01-29
[WIP] Instance-level configuration

flash-gordon:
* This adds support for instance-level configuring. For that, you use `include
Dry::Configurable` rather than `extend`.
* Internal NONE value was replaced with Undefined from dry-core (also dry-core is a dep
  now).
* ArgumentParser was reworked, it now handles processor, rejects unknown options
and always returns default options.

### DIFF
--- a/dry-configurable.gemspec
+++ b/dry-configurable.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.3.0"
   spec.add_runtime_dependency 'concurrent-ruby', '~> 1.0'
+  spec.add_runtime_dependency 'dry-core', '~> 0.4', '>= 0.4.7'
 
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'rake'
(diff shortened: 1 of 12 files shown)


timriley:
@flash-gordon This looks all fair to me.

When I was talking to you about this before, I was suggesting that a logic next step from here may be to allow dry-configurable to allow users to actually build up and customise their own config classes.

For example:

```ruby
class MyConfig < Dry::Configurable::Config # whether this is a subclassing thing or a mixin, I dunno
  setting :foo
  setting :bar
end

my_config = MyConfig.new(foo: "foo value", bar: "bar value")
```

Basically, kind of like a specialised struct builder built with the intention and features related to holding config (i.e. everything we can currently do with the `setting` API should be able to do here too, but the settings relate to _instance attributes_)

This would perhaps allow classes like https://github.com/hanami/hanami/blob/unstable/lib/hanami/configuration.rb to be built up without massive amounts of boilerplate. And because these are just ordinary user-controlled classes now, it would open up some potentially interesting uses. Either way, what it definitely would open up is dry-configurable supporting the case of _injected config objects_, which it cannot really do now.

Does that make sense to you?

timriley:
@flash-gordon anyway, the reason I mention this here is because if it sounds like something we like (👋 @solnic too), it'd just be good to make sure we can logically build this into a future release without having to break APIs unnecessarily (my feeling is it _should_ be OK, but it certainly never hurts to get a second opinion!)

PR #65, opened 2019-02-20
fix NoMethodError when reset module config

Neznauy:
Hello,
There is  ```NoMethodError: undefined method `_settings'``` when trying to perform ```reset_config``` method to module, extended by ```Dry::Configurable```.
Seems like it happens because of module doesn't pass check ```self.is_a?(Class)``` (because it is a ```Module``` class instead of ```Class``` class) and goes to ```else``` block, and ```_settings``` method applied to ruby ```Module``` class, which doesn't know about this method.
So, this PR fix this.

### DIFF
--- a/lib/dry/configurable/test_interface.rb
+++ b/lib/dry/configurable/test_interface.rb
@@ -8,7 +8,7 @@ module Dry
       #
       # @api public
       def reset_config
-        @config = if [Class, Module].include? self.class
+        @config = if self.is_a?(Module)
                     _settings.create_config
                   else
                     self.class._settings.create_config
--- a/lib/dry/configurable/test_interface.rb
+++ b/lib/dry/configurable/test_interface.rb
@@ -8,11 +8,11 @@ module Dry
       #
       # @api public
       def reset_config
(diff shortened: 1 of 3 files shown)


Neznauy:
@flash-gordon

lib/dry/configurable/test_interface.rb
```diff
@@ -8,11 +8,11 @@ module TestInterface
       #
       # @api public
       def reset_config
-        if self.is_a?(Class)
-          @config = _settings.create_config
-        else
-          @config = self.class._settings.create_config
-        end
+        @config = if [Class, Module].include? self.class
```
flash-gordon: `Module` is a superclass of `Class` so you only need to change it to `is_a?(Module)`, should be enough

flash-gordon:
I didn't think about this case. Obviously, because it wasn't covered in the specs, thanks for adding them.

lib/dry/configurable/test_interface.rb
Neznauy: yes, this is better

PR #83, opened 2020-02-21
Store settings as symbol and fetch them from symbol or string

### LINKED ISSUE #82: Accessing settings as strings via `#[]`
**Describe the bug**

Before the update to 0.11, a setting could be referenced via a string through `config.[]`. Now it must be a symbol, otherwise the setting is not found.

I'd like to know if it's desired behavior. If not, I can send a patch myself.

**To Reproduce**

```ruby
class A
  extend Dry::Configurable

  setting :foo, :bar
end

A.config[:foo]
# => :bar
A.config["foo"]
# => ArgumentError (+foo+ is not a setting name)
```

**Expected behavior**

```ruby
A.config["foo"]
# => :bar
```

**Your environment**

Affects two gems I maintain:

https://github.com/waiting-for-dev/warden-jwt_auth/issues/21
https://github.com/waiting-for-dev/devise-jwt/issues/159

waiting-for-dev:
Closes #82

I also added the counter part: always coercing name to symbol when storing a setting. Otherwise fetching a string setting would fail if a symbol were provided.

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -33,6 +33,7 @@ module Dry
       #
       # @return Config value
       def [](name)
+        name = name.to_sym
         raise ArgumentError, "+#{name}+ is not a setting name" unless _settings.key?(name)
 
         _settings[name].value
--- a/lib/dry/configurable/dsl.rb
+++ b/lib/dry/configurable/dsl.rb
@@ -43,7 +43,7 @@ module Dry
 
         default, opts = args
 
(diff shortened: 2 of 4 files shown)


waiting-for-dev:
Thanks for the quick review and approval @solnic Could I have a release with this change so that I can fix my two libraries?

solnic:
@waiting-for-dev yes I will release it today

waiting-for-dev:
Thanks a lot @solnic

PR #84, opened 2020-02-23
WIP Forwarding setting with default value

waiting-for-dev:
In relation with a [topic discussed in the chat](https://dry-rb.zulipchat.com/#narrow/stream/191662-general/topic/dry-configurable/near/188869250). @solnic I think I've managed to isolate the main problem I'm having.

This WIP PR has a failing test which should be working before last major update. When the setting in `Glass` has no pre-processor then it works, but when it has one it seems it's not called when setting from `Klass`:

```
 1) Dry::Configurable::Config#forwarding can forward setting to another class
     Failure/Error: value.upcase
     
     NoMethodError:
       undefined method `upcase' for nil:NilClass
```

Any idea what it's going on?

Thanks

### DIFF
--- a/spec/unit/dry/configurable/dsl_spec.rb
+++ b/spec/unit/dry/configurable/dsl_spec.rb
@@ -49,6 +49,14 @@ RSpec.describe Dry::Configurable::DSL do
     expect(setting.value).to eql('jdbc:sqlite')
   end
 
+  it 'compiles a setting with a constructor and no default value' do
+    setting = dsl.setting(:name) { |value| value.upcase }
+
+    expect(setting.name).to be(:name)
+    # What should we expect for setting.value?
+    expect(setting.value).to eql('jdbc:')
+  end
+
   it 'compiles a nested list of settings' do
     setting =
       dsl.setting(:db) do


solnic:
@waiting-for-dev no I don’t. Could you debug that?

waiting-for-dev:
@solnic it turned out the underlying problem doesn't require such a complicated setup. I updated this PR so that you can see the problem. It seems that pre-processors are eagerly called when configuring the setting. The problem is that when we don't have a default value for the setting it can crash if the pre-processor tries to call a method on the value (which turns out to be `nil`). Is it a decision by design? In the test in this PR, what `setting.value` should return in this case?

timriley:
@waiting-for-dev I've hit this very same issue too, I think.

I played around a bit and made something work for my _particular_ use case, but it breaks some important tests: https://github.com/dry-rb/dry-configurable/pull/87

I'm still thinking through a way to solve this... I'll be sure to keep you in the loop!

timriley:
This should be fixed by #87 now. @waiting-for-dev, if you have any further issues, please do re-open this or create another issue!

PR #86, opened 2020-03-01
Use keyword arguments for all setting options

timriley:
This PR adjusts `setting` parameters such that `default:` and `constructor:` are now keyword params.

This clarifies usage and simplifies internals.

The expected arguments are the same as before, with the acceptable range of types for `constructor:` becoming even broader, since it can now accept any callable object, instead of the singular block argument that Ruby methods can take.

Since this is a very widely used gem, the previous API is still supported:

- `default` as the second positional parameter (after setting name)
- constructor/processor as a block argument with arity >0

When either of these are used, we output a deprecation message to warn that these parameters will not be supported in the next "major" version.


timriley:
@solnic @flash-gordon Do you want to go through a deprecation cycle for the old API (with the `default` passed as a positional param and the `constructor` passed as a block)?

If so, I'd be happy to add that in.

timriley:
I've answered this question for myself: given the number of other gems in our ecosystem that depend on dry-configurable, I think we _definitely_ need to run this change through a full API deprecation cycle

timriley:
@solnic @flash-gordon Ready for you to review!

FWIW I do plan to follow this up with the one other behavioural change we discussed re: constructors – that they be passed `Undefined` if the setting is evaluated without a default or input value assigned. That'll likely be next week.

solnic:
@timriley seems like DSL coverage dropped because deprecated scenarios are...not covered? which is weird, is this simplecov not doing its thing correctly?

timriley:
@solnic I think it's actually right — let me add some tests now to explicitly cover the deprecated scenarios 😬

timriley:
@solnic sorted!

solnic:
@timriley unfortunately this is not backward compatible. dry-schema broke 😢 I recommend leaving specs as-they-are and adding new specs that would cover the updated API. I suspect that this will uncover what broke. Digging into dry-schema failures is going to be much more difficult.

timriley:
@solnic What broke, exactly? I just ran dry-schema's whole test suite on this branch of dry-configurable and it fully passed.

timriley:
dry-validation passed too.

I did find one breakage in dry-system's specs, though:

```ruby
            setting :log_levels,
              development: Logger::DEBUG,
              test: Logger::DEBUG,
              production: Logger::ERROR
```

Because we're now accepting `**kwargs`, this default value of a hash is now being seen as 3 different keyword arguments 😞

solnic:
> Because we're now accepting **kwargs, this default value of a hash is now being seen as 3 different keyword arguments 😞

that's a tough one, I start to think we may want to just break it and release 1.0.0 already :/

timriley:
@solnic Yeah, it doesn't feel like there's an easy way to support that usage from dry-system's logging plugin.

So yeah, I'm ok for all of this just to go into 1.0, at least that way we don't potentially mess up people's apps since we'll be clearly upping the major version.

We'll probably want to bunch up all the work we need for 1.0 so this effort doesn't stretch out too long!

flash-gordon (approved):
somehow I missed it but it looks good :)

solnic:
oops, that last commit was by an accident, I'll remove it

timriley:
BTW, this is on hold from my end until I get back to making Hanami's application setting use dry-configurable. At that point, I'll do the final batch of pre-1.0 breaking changes and we can test them via some betas.

waiting-for-dev:
This can be closed as the work has been integrated into:

- https://github.com/dry-rb/dry-configurable/pull/112
- https://github.com/dry-rb/dry-configurable/pull/111

solnic:
Can we delete the branch too?

waiting-for-dev:
I think we can, but as it's @timriley's work, I prefer to wait for his consent.

PR #87, opened 2020-03-01
Do not clone unevaluated values when copying Settings

timriley:
This is important for working with settings that have a constructor, but have not been provided an input yet.

This makes it possible to build settings with constructors expecting certain types of values, and then progressively apply those values across a range of settings on an object before finalizing it.

### DIFF
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -63,8 +63,18 @@ module Dry
       end
 
       # @api private
+      def input_defined?
+        !input.equal?(Undefined)
+      end
+
+      # @api private
       def value
         @value ||= evaluate
+      end
+
+      # @api private
+      def evaluated?
+        instance_variable_defined?(:@value)
       end
 
       # @api private
@@ -107,7 +117,7 @@ module Dry
       # @api private
       def initialize_copy(source)
         super
-        @value = source.value.dup if source.clonable_value?
+        @value = source.value.dup if source.input_defined? && source.clonable_value?
         @options = source.options.dup
       end
 
--- a/spec/unit/dry/configurable/setting_spec.rb
+++ b/spec/unit/dry/configurable/setting_spec.rb
@@ -8,6 +8,38 @@ RSpec.describe Dry::Configurable::Setting do
     Dry::Configurable::Setting.new(:test, **options)
   end
 
+  describe '#input_defined?' do
+    context 'no input provided' do
+      let(:options) do
+        {}
+      end
+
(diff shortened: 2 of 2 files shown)


solnic:
> Do you have any thoughts about how to keep something like this in place while still having the other specs pass?

Looks like this change broke inheritance. ie [this spec](https://github.com/dry-rb/dry-configurable/blob/adac57bd91f7eb6f69b52e5a2a113e5da9617ec1/spec/integration/dry/configurable/config_spec.rb#L81) failure is legit. What happens is that during inheritance the parent doesn't have values evaluated, so its defaults are not copied for the child.

We need to somehow track when a value must be cloned for inheritance despite the fact it wasn't evaluated yet.

timriley:
@solnic yeah, this change has definitely broken behaviour — I wanted to get something up to prompt a bit of discussion about it.

I think perhaps part of the approach to solve this might be to evaluate the _default_ (if there is one) at the time of cloning. That's what I was going to look at next, anyway.

timriley:
@solnic @flash-gordon — OK, how about this? Tests are all green again _and_ this satisfies the behavioural change I needed for Hanami (being able to iterate through settings, some with constructor process, and apply values to each progressively).

I'll look to add a test to exercise this behaviour next, but just flagging it with you to get your feedback in the meantime.

lib/dry/configurable/setting.rb
```diff
@@ -62,6 +62,11 @@ def initialize(name, input: Undefined, default: Undefined, **options)
         @options = options
       end
 
+      # @api private
+      def input_defined?
+        input != Undefined
```
solnic:
```suggestion
        !input.equal?(Undefined)
```

waiting-for-dev:
Hey, sorry, I wanted to help with it but I've been very busy last weeks... As far as I see, there's only a very minimal suggestion and we could fix it, isn't it? Please, let me know if I can help with anything

timriley:
Thanks for the nudge here @waiting-for-dev, I’ll sort this out in this coming week.

timriley:
@solnic @flash-gordon @waiting-for-dev 

OK, this is ready to go.

I've expanded the unit tests for `Setting` to cover this new behaviour, and through doing that, added `#evaluated?` back onto `Setting` since it seems like a useful thing to have – at very least for the tests to effectively describe the state of a setting.

Let me know if you're good and I can merge and then cut a new minor release.

waiting-for-dev:
Thanks a lot @timriley :smile:

PR #97, opened 2020-08-04
Do not memoize config when using pre-process values

### LINKED ISSUE #96: Values are being memoized between config instances
**Describe the bug**

This is a regression between versions 0.11.5 and 0.11.6. When you instantiate multiple instances of a class that include `Dry::Configurable`, the default values specified in the `setting` block are memoized between the instances. 

I'm including a simplified reproduction script based on an issue with https://github.com/karafka/waterdrop/.

You can instantiate multiple WaterDrop consumers, each with different settings. WaterDrop has a default `monitor` (based on `dry-monitor`), which is instantiated by `dry-configurable`. 

We subscribe to each consumer's monitor so that we can publish instrumentation metrics:

```ruby
listener = KafkaStatsdListener.new
publisher.monitor.subscribe(listener)
publisher2.monitor.subscribe(listener)
```

This works fine with 0.11.5, but with 0.11.6 the messages from the monitor are duplicated.

**To Reproduce**

```ruby
require 'dry-configurable'
require 'dry/configurable/version'

MyMonitor = Class.new

class Producer
  include Dry::Configurable

  setting(:monitor, false) { |monitor| monitor || MyMonitor.new }
end

producer1 = Producer.new
producer2 = Producer.new

puts "dry-configurable version: #{Dry::Configurable::VERSION}"
puts producer1.config.monitor.object_id == producer2.config.monitor.object_id
```

Output:
```
dry-configurable version: 0.11.6
true
```

**Expected behavior**

```
dry-configurable version: 0.11.5
false
```

esparta:
There's an corner case when a configuration should not be memoized
when using pre-processor values:

```ruby
MyMonitor = Class.new

class Producer
  include Dry::Configurable
  setting(:monitor, false) { |monitor| monitor || MyMonitor.new }
end

p01 = Producer.new
p02 = Producer.new

raise 'Not memoized allowed' \
    if p01.config.monitor.object_id == p01.config.monitor.object_id
```

This behavior was introduced on Dry::Config 0.11.6, on PR #95

Fixed #96

Notes: I'm modifying the tests for `Dry::Config::Setting` because the test
was interfeiring with the behavior when introduce a constructor which
happend to be exactly the issue when using the setting method.

### DIFF
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -61,7 +61,7 @@ module Dry
         @default = default
         @options = options
 
-        evaluate if input_defined?
+        evaluate if input_defined? && !options[:constructor].is_a?(Proc)
       end
 
       # @api private
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -60,8 +60,6 @@ module Dry
         @input = input.equal?(Undefined) ? default : input
         @default = default
         @options = options
-
-        evaluate if input_defined? && !options[:constructor].is_a?(Proc)
       end
 
       # @api private
@@ -114,6 +112,11 @@ module Dry
         CLONABLE_VALUE_TYPES.any? { |type| value.is_a?(type) }
       end
 
+      # @api private
+      def evaluate
+        @value = constructor[input.equal?(Undefined) ? nil : input]
+      end
+
       private
 
       # @api private
@@ -121,11 +124,6 @@ module Dry
         super
         @value = source.value.dup if source.input_defined? && source.clonable_value?
         @options = source.options.dup
-      end
-
-      # @api private
-      def evaluate
-        @value = constructor[input.equal?(Undefined) ? nil : input]
       end
     end
   end
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -108,7 +108,9 @@ module Dry
         if setting.writer?(meth)
           raise FrozenConfig, 'Cannot modify frozen config' if frozen?
 
-          _settings << setting.with(input: args[0])
+          new_setting = setting.with(input: args[0])
+          new_setting.evaluate
+          _settings << new_setting
         else
           setting.value
         end
--- a/spec/integration/dry/configurable/setting_spec.rb
+++ b/spec/integration/dry/configurable/setting_spec.rb
@@ -162,6 +162,27 @@ RSpec.describe Dry::Configurable, '.setting' do
         expect { klass.setting name }.to raise_error(ArgumentError, /not a valid/)
       end
     end
+
+    context 'allow immediate feedback on failable settings' do
+      before do
+        klass.setting :failable do |value|
+          value.to_sym unless value.nil?
+        end
+      end
+
+      it 'fail if using bad input with .config.setting=' do
+        expect do
+          object.config.failable = 12
+        end.to raise_error(NoMethodError, /undefined method `to_sym'/)
+      end
+      it 'assign input with .configure DSL' do
+        expect do
+          object.configure do |config|
+            config.failable = 12
+          end
+        end.to raise_error(NoMethodError, /undefined method `to_sym'/)
+      end
+    end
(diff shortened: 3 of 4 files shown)


solnic:
@esparta thanks for tackling this! I checked it against schema/validation/system/rails test suites and they all pass. @timriley could you see if this looks good?

lib/dry/configurable/setting.rb
```diff
@@ -61,7 +61,7 @@ def initialize(name, input: Undefined, default: Undefined, **options)
         @default = default
         @options = options
 
-        evaluate if input_defined?
+        evaluate if input_defined? && !options[:constructor].is_a?(Proc)
```
esparta, line 64: This can be simplified using the truthy value

```suggestion
        evaluate if input_defined? && !options[:constructor]
```

I added a reviewing to also be a `Proc` just to explain better the intention, but can be removed easyly.

timriley:
I’ll try check this on the weekend!

timriley:
Hi folks, I've just looped back to this, and I think this change won't preserve the behaviour I introduced in #95. I think to be sure about our way forward here, we want to add at least 2 higher-level use-case specs expressing our desired behaviours, both the one in #95 as well as the desired (non buggy) behaviour in #96.

Then they can ensure that whatever code changes we make still satisfy those two desired use cases.

@esparta Here's the use case I was wanting to satisfy:

Given a setting like this:

```ruby
setting :default_request_format do |value|
  Hanami::Utils::Kernel.Symbol(value) unless value.nil?
end
```

I want this behaviour:

```ruby
expect {
  configuration.default_request_format = 23
}.to raise_error(TypeError)
```

Did you want to have a go setting up those use case specs as a way to drive our work here?

timriley (changes requested):
Requesting that we go ahead with the high-level use case specs as a way to verify we're properly satisfying both sets of requirements here.

esparta:
So I change approach to have those high-level cases covered:

- Removing the caching on pre-process values
- Provide immediate response on failable settings

Instead of having the logic in `settings`, the evaluation is done right after the new setting is created on reassignment (config.rb).
I added a new integration test having a failable setting with immediate response for config errors.

Please let me know if this new approach is acceptable & complete for the specs, @timriley

lib/dry/configurable/setting.rb
```diff
@@ -114,6 +112,11 @@ def clonable_value?
         CLONABLE_VALUE_TYPES.any? { |type| value.is_a?(type) }
       end
 
+      # @api private
+      def evaluate
```
esparta, line 116: I did change the scope to public, in order to have the evaluation available at config.rb

lib/dry/configurable/setting.rb
```diff
@@ -60,8 +60,6 @@ def initialize(name, input: Undefined, default: Undefined, **options)
         @input = input.equal?(Undefined) ? default : input
         @default = default
         @options = options
-
-        evaluate if input_defined?
```
esparta, line 64: I did consider to have a new parameter to evaluate conditionally, but IMO this class already has too many parameters.
The evaluation is only needed in one place, so removing this conditional and do the evaluation only when needed seems like a better fit for the requirement.

timriley:
Thanks for setting this up, @esparta! I'm just doing a few investigations into alternative approaches; I'll get back to you :)

timriley:
@esparta @solnic So I figured out another approach to this that allows the fix to be fully contained within `Setting`, rather than spread over both `Setting` and `Config` (along with a private `Setting` method being made public), which is what felt uncomfortable about this approach here.

It's over in https://github.com/dry-rb/dry-configurable/pull/99 — would you mind taking a look and letting me know what you think?

Thanks so much for putting forward a first cut at the fix here, @esparta, and for writing those extra tests! 🙏 Without this effort I'm pretty sure we wouldn't have go this far :)

esparta:
No problem, I'm all in having the solution all contained in `Setting`, also I didn't feel comfortable changing the scope of evaluate.
We can move the conversation to #99 and close this PR.

PR #100, opened 2021-01-04
Support custom cloneable values

timriley:
With this change, setting value objects responding to `#cloneable_value?` and returning true will be cloned when settings are cloned or dup’ed, in addition to the existing fixed set of classes (Array, Hash, Set, Config).

This makes it possible to provide “rich” config values that carry their own configuration interface, e.g.

```ruby
setting :component_dirs, Configuration::ComponentDirs.new
```

In this case, `ComponentDirs` could provide its own API for adding component dirs and configuring aspects of their behavior at the same time. Provided it implements its own appropriate `#initialize_copy` and then returns true to `#cloneable_value?`, dry-configurable will ensure that this value is cloned along its setting at the appropriate times.

As the example above indicates, this change is in service of dry-system providing a new ”rich” `component_dirs` setting, providing an API like this:

```ruby
# Inside a Dry::System::Container subclass
configure do |config|
  config.component_dirs.add("path/to/dir") do |dir|
    dir.auto_register = true
    dir.add_to_load_path = true
    # etc.
  end
end
```

### DIFF
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -19,7 +19,7 @@ module Dry
 
       DEFAULT_CONSTRUCTOR = -> v { v }.freeze
 
-      CLONABLE_VALUE_TYPES = [Array, Hash, Set, Config].freeze
+      CLONEABLE_VALUE_TYPES = [Array, Hash, Set, Config].freeze
 
       # @api private
       attr_reader :name
@@ -54,8 +54,8 @@ module Dry
       end
 
       # @api private
-      def self.clonable_value?(value)
-        CLONABLE_VALUE_TYPES.any? { |type| value.is_a?(type) }
+      def self.cloneable_value?(value)
+        CLONEABLE_VALUE_TYPES.any? { |type| value.is_a?(type) }
       end
 
       # @api private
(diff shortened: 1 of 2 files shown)


solnic (commented):
I have mixed feelings about the `cloneable_value?` approach - why not have a way of configuring which value types should be cloned? Right now it feels backwards, it is the config object that should know which types are cloneable rather than cloneable objects knowing that they are cloneable, especially that it's a configurable-specific knowledge, nothing generic.

timriley:
@solnic Fair points! So you're thinking something like this?

```
setting :component_dirs, MyCustomThing.new, cloneable: true
```

solnic:
> @solnic Fair points! So you're thinking something like this?
> 
> 
> 
> ```
> 
> setting :component_dirs, MyCustomThing.new, cloneable: true
> 
> ```

Pretty much, yes

timriley:
@solnic I'll take another pass in that direction :)

flash-gordon:
Exactly same thoughts came into mind before reading @solnic's comment. Otherwise, we can go as far as `setting obj` where `obj` has `setting_name`, `clonable_value?` and whatnot.
I'd say a more general approach is
```ruby
def setting(name, value, clonable: false, clone_with: Undefined)
  if clonable
    clone = Undefined.default(clone_with) { :initialize_copy.to_proc }
  else
    clone = Dry::Core::Constants::IDENTITY
  end
  ...
```
Not saying it's needed but we can keep it in mind.

timriley:
Closing in favour of #102.

solnic:
@timriley is it OK to delete the branch?

timriley:
Deleted! Thanks for the prompt @solnic :)

PR #101, opened 2021-01-04
Attempt to call super in finalize! method

adam12:
The `Dry::Configurable::Methods` module ends up in the ancestory
tree, but fails to call `super`.

This breaks scenarios where `finalize!` has been defined in a
configurable class, since the `finalize!` method in that class is the
third in the tree and is never reached.

Fixes https://github.com/hanami/view/issues/184

### DIFF
--- a/lib/dry/configurable/methods.rb
+++ b/lib/dry/configurable/methods.rb
@@ -25,6 +25,9 @@ module Dry
         return self if config.frozen?
 
         config.finalize!
+
+        super if defined?(super)
+
         self
       end
     end
--- a/spec/integration/dry/configurable/included_spec.rb
+++ b/spec/integration/dry/configurable/included_spec.rb
@@ -45,4 +45,23 @@ RSpec.describe Dry::Configurable, '.included' do
 
     it_behaves_like 'configure'
   end
+
+  context 'when #finalize! is defined in configurable class' do
+    let(:instance) do
+      Class.new do
+        include Dry::Configurable
+
+        attr_accessor :finalized
+
+        def finalize!
+          @finalized = true
+        end
+      end.new
+    end
+
+    it 'calls finalize! in configurable class' do
+      instance.finalize!
+      expect(instance.finalized).to be(true)
+    end
+  end
 end


solnic:
@adam12 thanks! I reckon this is behavior we want, right @timriley? I'm asking because I have a vague memory of discussing whether `finalize!` should be here after all but I *think* we want it after all 🤔

jodosha:
This also fixes `hanami-controller` `unstable` branch. I had to [pin](https://github.com/hanami/controller/blob/unstable/Gemfile#L29) `dry-configurable` to `0.11.6` to avoid this problem.

solnic:
@jodosha what is it that `hanami-controller` needs?

timriley:
Thanks so much for tracking this down and proposing a fix, @adam12!

I'm not sure I'm fully happy with the change in #98, which led to this issue in the first place.

I would consider it unexpected behaviour for dry-configurable to be prepending all of its instance methods (i.e. the exact change that was made in #98). My policy on prepending is that it should be the exception, rather than the rule, and when used, it should be for an obvious requirement for the library's behaviour, and and clearly communicated as part of the library's documentation.

Prepending _all_ the configurable instance methods I would call a violation of the principal of least surprise. Prepending simply isn't needed for most of them, and we have a clear example of this: it became a surprising bug for both hanami-view and hanami-controller!

Looking over #98 more deeply, it seems there could be a few other ways of solving the core issue:

1. Simply document that when including Dry::Configurable, any custom initialize method should call super as its first first step. This is not super nice, and it would be best if we could avoid it.
2. Move initialisation of the `@config` ivar into its own memoized reader method, making it lazy. This could, however, raise some other behavioural issues if any other parts of the gem rely on the config being initialized immediately after an object with instance-based configuration is initialized. Might be worth a quick look, though
3. Create an additional `Initializer` module, separate to the `InstanceMethods` module, move the `#initialize` method into this module, and then _prepend this module only,_ leaving the `InstanceMethods` to be included like before

All of these would address the base issue without us having to do any workarounds like `super if defined?(super)`, which for most sorts of library code, I would consider a smell, and a sign that other parts of the library should be restructures such that it isn't necessary.

What do you think, @adam12, @solnic, @flash-gordon?

solnic:
@timriley thanks for such a detailed explanation and I'm sorry I rushed with merging the aforementioned PR. Re proposed solutions, I've a strong opinion it should be opt 3. The way I see it, is that in cases like this one, the user of a lib should not be bothered by details like when something happens. Things should Just Work™. That's why forcing people to call `super` is just extra friction in the UX that we could avoid here. There are of course cases where this is unavoidable, but it seems like here we could avoid it.

flash-gordon:
I'm inclined to 3. Although personally, I don't see a problem in "forcing" people to call `super`. Knowing how a language or library works shouldn't harm. We could have a good error message when `@config` is missing. But I wish ruby has it out of the box making sure you call `super` in constructors.

jodosha:
I'm for 3 as well 😃

adam12:
Thanks folks. I've opened #103 to implement option 3.

PR #102, opened 2021-01-12
Add `cloneable` setting option

timriley:
This is an evolution from the approach in #100. With this change, settings may be specified with a `cloneable` option, e.g.

```ruby
setting :component_dirs, Configuration::ComponentDirs.new, cloneable: true
```

This makes it possible to provide “rich” config values that carry their own configuration interface.

In the example above, `ComponentDirs` could provide its own API for adding component dirs and configuring aspects of their behavior at the same time. By being passed to the setting along with `cloneable: true`, dry-configurable will ensure the setting's values are cloned along with the setting at all the appropriate times (see `Setting#initialize_copy`).

As the example here suggests, this change is in service of dry-system providing a new ”rich” `component_dirs` setting, providing an API like this:

```ruby
# Inside a Dry::System::Container subclass
configure do |config|
  config.component_dirs.add("path/to/dir") do |dir|
    dir.auto_register = true
    dir.add_to_load_path = true
    # etc.
  end
end
```

### DIFF
--- a/changelog.yml
+++ b/changelog.yml
@@ -1,4 +1,25 @@
 ---
+- version: Unreleased
+  summary:
+  date:
+  added:
+  - |-
+    Settings may be specified with a `cloneable` option, e.g.
+
+    ```ruby
+    setting :component_dirs, Configuration::ComponentDirs.new, cloneable: true
+    ```
+
+    This change makes it possible to provide “rich” config values that carry their own
+    configuration interface.
+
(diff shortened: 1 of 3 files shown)


timriley:
FYI, I've tested this in https://github.com/dry-rb/dry-system/pull/155 and it's working nicely there.

flash-gordon (approved):

solnic (approved):

PR #103, opened 2021-01-14
Prepend initialize method only

adam12:
Move initialize method to it's own module and prepend
it inside the `include` hook.

Revert the prepending of `InstanceMethods` module in
favour of using `include`.

Fixes instances where `finalize!` methods in classes that
include `Dry::Configurable` are never called due to the
ancestor chain.

Fixes hanami/view#184


solnic (approved):
Thanks for taking care of this. 👍🏻 from me.

PR #105, opened 2021-04-10
Make `Config#finalize!` finalize itself recurively and allow freezing values

### LINKED ISSUE #104: `finalize!` should freeze config recursively
## Describe the bug

`#finalize!` freezes the main config, but internal configs are still not frozen.

## To Reproduce

```
require 'dry-configurable'

class App
  extend Dry::Configurable

  setting :database do
    setting :dsn, 'sqlite:memory'
  end
end

App.config.finalize!

App.config.database.dsn = 'wrong!'
p App.config.database.dsn # => wrong!
App.config.database = 'wrong!' # => Cannot modify frozen config
```

## Expected behavior

`App.config` is immutable, so even nested values can't be changed.

## My environment

- Affects my production application: Maybe?
- Ruby version: 3.0.0
- OS: linux x86_64
- dry-configurable (0.12.1)

ojab:
Fixes #104

### DIFF
--- a/lib/dry/configurable/instance_methods.rb
+++ b/lib/dry/configurable/instance_methods.rb
@@ -35,8 +35,6 @@ module Dry
       #
       # @api public
       def finalize!
-        return self if frozen?
-
         super
         freeze
       end
--- a/lib/dry/configurable/instance_methods.rb
+++ b/lib/dry/configurable/instance_methods.rb
@@ -34,7 +34,7 @@ module Dry
       # Finalize the config and freeze the object
       #
       # @api public
-      def finalize!
+      def finalize!(freeze_values: false)
         super
         freeze
       end
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -66,7 +66,7 @@ module Dry
         @default = default
         @options = options
 
-        value if input_defined?
+        evaluate if input_defined?
       end
 
       # @api private
@@ -80,6 +80,8 @@ module Dry
 
         @value = constructor[Undefined.coalesce(input, default, nil)]
       end
+      alias_method :evaluate, :value
+      private :evaluate
 
       # @api private
       def evaluated?
@@ -97,8 +99,12 @@ module Dry
       end
 
       # @api private
-      def finalize!
-        value.finalize! if value.is_a?(Config)
+      def finalize!(freeze_values: false)
+        if value.is_a?(Config)
+          value.finalize!(freeze_values: freeze_values)
+        elsif freeze_values
+          value.freeze
+        end
         freeze
       end
 
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -66,7 +66,7 @@ module Dry
         @default = default
         @options = options
 
-        evaluate if input_defined?
+        value if input_defined?
       end
 
       # @api private
@@ -76,7 +76,9 @@ module Dry
 
       # @api private
       def value
-        @value ||= evaluate
+        return @value if evaluated?
+
+        @value = constructor[Undefined.coalesce(input, default, nil)]
       end
 
       # @api private
@@ -92,6 +94,12 @@ module Dry
       # @api private
       def pristine
         with(input: Undefined)
+      end
+
+      # @api private
+      def finalize!
+        value.finalize! if value.is_a?(Config)
+        freeze
       end
 
       # @api private
@@ -140,11 +148,6 @@ module Dry
           @default = source.default.dup
           @value = source.value.dup if source.evaluated?
         end
-      end
-
-      # @api private
-      def evaluate
-        @value = constructor[Undefined.coalesce(input, default, nil)]
       end
     end
   end
--- a/spec/integration/dry/configurable/setting_spec.rb
+++ b/spec/integration/dry/configurable/setting_spec.rb
@@ -392,6 +392,23 @@ RSpec.describe Dry::Configurable, ".setting" do
       expect { object.configure {} }.to raise_error(Dry::Configurable::FrozenConfig)
     end
 
+    it "can be finalized with freezing values" do
+      klass.setting :kafka, "kafka://127.0.0.1:9092"
+
+      object.finalize!(freeze_values: true)
+      # becomes a no-op
+      object.finalize!(freeze_values: true)
+
+      expect(object).to be_frozen
+      expect(object.config.db).to be_frozen
+      expect(object.config.db.user).to be_frozen
+
+      expect { object.config.db.user << "foo" }.to raise_error(FrozenError)
+
+      # does not allow configure block anymore
+      expect { object.configure {} }.to raise_error(Dry::Configurable::FrozenConfig)
+    end
+
     it "defines a reader shortcut for nested config" do
       klass.setting :dsn, reader: true do
         setting :pool, default: 5
--- a/spec/integration/dry/configurable/setting_spec.rb
+++ b/spec/integration/dry/configurable/setting_spec.rb
@@ -328,7 +328,7 @@ RSpec.describe Dry::Configurable, ".setting" do
         klass.setting :env
 
         klass.setting :db do
-          setting :user, default: "root"
+          setting :user, default: "root".dup
           setting :pass, default: "secret"
         end
       end
@@ -375,13 +375,18 @@ RSpec.describe Dry::Configurable, ".setting" do
     end
 
     it "can be finalized" do
-      klass.setting :db, default: "sqlite"
+      klass.setting :kafka, default: "kafka://127.0.0.1:9092"
 
       object.finalize!
       # becomes a no-op
       object.finalize!
 
       expect(object).to be_frozen
+      expect(object.config.db).to be_frozen
+      expect(object.config.db.user).not_to be_frozen
+
+      object.config.db.user << "foo"
+      expect(object.config.db.user).to eq("rootfoo")
 
       # does not allow configure block anymore
       expect { object.configure {} }.to raise_error(Dry::Configurable::FrozenConfig)
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -82,7 +82,7 @@ module Dry
 
       # @api private
       def finalize!
-        _settings.freeze
+        _settings.finalize!
         freeze
       end
 
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -81,8 +81,8 @@ module Dry
       alias_method :to_h, :values
 
       # @api private
-      def finalize!
-        _settings.finalize!
+      def finalize!(freeze_values: false)
+        _settings.finalize!(freeze_values: freeze_values)
         freeze
       end
 
--- a/lib/dry/configurable/methods.rb
+++ b/lib/dry/configurable/methods.rb
@@ -21,8 +21,8 @@ module Dry
       # @return [Dry::Configurable::Config]
       #
       # @api public
-      def finalize!
-        config.finalize!
+      def finalize!(freeze_values: false)
+        config.finalize!(freeze_values: freeze_values)
         self
       end
     end
--- a/lib/dry/configurable/methods.rb
+++ b/lib/dry/configurable/methods.rb
@@ -22,8 +22,6 @@ module Dry
       #
       # @api public
       def finalize!
-        return self if config.frozen?
-
         config.finalize!
         self
       end
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -54,6 +54,12 @@ module Dry
         self.class.new(map(&:pristine))
       end
 
+      # @api private
+      def finalize!
+        each(&:finalize!)
+        freeze
+      end
+
       private
 
       # @api private
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -55,8 +55,8 @@ module Dry
       end
 
       # @api private
-      def finalize!
-        each(&:finalize!)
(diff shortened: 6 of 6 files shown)


spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -377,13 +377,13 @@
     end
 
     it 'can be finalized' do
-      klass.setting :db, 'sqlite'
```
solnic, line 380: Why did you have to remove this line?

solnic (commented):
Thanks for tackling this! I left one comment. In general, we're trying to wrap up dry-configurable 1.0.0 and since it's used in a bunch of projects (+ we want to use it in hanami 2.0) we need to 100000% sure we're not breaking anything. Some behaviors are really tricky so we probably have to check your branch against dry-schema and dry-system too before merging.

lib/dry/configurable/setting.rb
```diff
@@ -94,6 +96,12 @@ def pristine
         with(input: Undefined)
       end
 
+      # @api private
+      def finalize!
+        value.is_a?(Config) ? value.finalize! : value.freeze
```
flash-gordon, line 101: I don't think that freezing values is the right way. You can't know in advance what values are, it can be anything. For instance, if it's a module or a class (a reasonable assumption for config!) then you won't be able to define new constants there. I don't this is expected.

flash-gordon (changes requested):

spec/integration/dry/configurable/setting_spec.rb
ojab: Thought that it's not needed, since we have a :db already, but on the second though we're testing that config could be modified. Put it back.

lib/dry/configurable/setting.rb
ojab: I thought that it would be questionable approach :)

My reasoning is that if we're finalizing config — it owns all the values and we should take that into account. IMHO
```
irb(main):005:0> App.config.database.dsn
=> "sqlite:memory"
irb(main):006:0> App.config.database.dsn << 'qwe'
=> "sqlite:memoryqwe"
```
is quite unexpected after we `#finalize!`, especially since documentation says `# Note that finalized configs cannot be changed.` & changelog says `Introduce `Configurable.finalize!` which freezes config and its dependencies`.

ojab:
Should I just check dry-schema/dry-system locally or make draft PRs to them using this branch?

spec/integration/dry/configurable/setting_spec.rb
ojab: Oh, right, we're overwriting nested `db: %i[user password]` here. Replaced with `klass.setting :kafka, 'kafka://127.0.0.1:9092'`

lib/dry/configurable/setting.rb
flash-gordon: Yeah, I understand the reasoning, I would be happy to have everything immutable in the first place. But my position on this it's not up to libraries to mutate objects provided by the user. Freezing is a type of mutation so this rule applies. There are two problems: 1) it's unexpected 2) you can't opt-out of this behavior. I don't think we should be unnecessarily smart about this. We could have a warning shown when a non-frozen value passed to the config but then I guess there'd be a ton of them in a typical project, I'm not sure if this would be helpful.

ojab: Right now semantics of `#finalize!` is not clearly defined, so what is the expected behavior? Freeze just config structure, so we're protected from deleting of used keys?
From the changelog entry I expected that `dependencies` are config values, not sure what was meant by `dependencies` if they're not.
Implicit mutation is certainly not a good thing, but `#finalize!` is explicit user's decision and if semantic would be defined as `freeze all the keys & values` — why not? If you want to opt-out — just don't call `#finalize!`.

If we still need to freeze the structure without values — can we change `#freeze` to freeze the config without values recursively, so the structure is set in stone, and make `#finalize!` to freeze also the values?

Right now `#finalize!` seems broken anyway, because freezing only top-level keys has doubtful usefullness.

flash-gordon: It's not that simple. `finalized!` is called by dry-system and it's an essinatial part in the lifecycle, you can't skip it. When people install new version of dry-system witth current change they'll get new semantics for the "why not?" reason. I agree it shouldn't break things but the problem is we can't be sure and our semantics will be different from what's understood as a default in ruby: calling `.freeze` on arrays and hashes doesn't freeze values. 

WRT recursive `finalize!` and freezing config structure I don't see problems with this. I agree it's not expected config structure can be changed after freezing.

solnic: Re dry-system, I believe it no longer depends on this hook (in master) but yeah, if the behavior changes here it'll break current dry-system releases.

solnic:
> Should I just check dry-schema/dry-system locally or make draft PRs to them using this branch?

Running locally should be enough. CI will later on run specs against dry-configuration master to double-check.

lib/dry/configurable/setting.rb
flash-gordon: Yeah, I checked, it's not called. It's not clear whether it should be though. All in all, I think we could have freezing values as an option `.finalize!(freeze_values: true)`. We can debate if it should be `true` or `false` by default but either way, it provides an opt-in/opt-out mechanism which is good. WDYT?

lib/dry/configurable/setting.rb
```diff
@@ -66,7 +66,7 @@ def initialize(name, input: Undefined, default: Undefined, **options)
         @default = default
         @options = options
 
-        evaluate if input_defined?
+        value if input_defined?
```
timriley, line 69: I can see why you've removed the private `#evaluate` method here, to keep all the logic in `#value` itself, but the net result is that this line here reads a lot worse – it's just a random noun hanging out there on its own line, rather than a verb describing the outcome of the method (which, in this case, is actually about a side effect, so a verb makes sense).

Is this just a change you just made along the way because you thought it would be nice, or is it critical to your work in this PR?

lib/dry/configurable/setting.rb
timriley: You're right that dry-system doesn't call `finalize!` on its config at the moment. This would be worth investigating before we ship 1.0. I'll file an issue.

In fact, I'm not sure if we really finalize configs much at all across the dry-rb gems at the moment.

Having a `freeze_values` to opt in or out of freezing non-`Config` values like a sensible way to go here 👍🏼

timriley: Issue for dry-system here: https://github.com/dry-rb/dry-system/issues/168

lib/dry/configurable/setting.rb
ojab: `@value ||=` instead of `evaluated?` leads to evaluating value during `cloneable?` check from `initialize_copy` which fails if `finalize!(freeze_values: true` is called. I don't quite like how it looks either, added alias evaluate -> value.

ojab:
Sorry for delay, changed it to `finalize!(freeze_values: false)`, checked dry-system & dry-schema from git, specs are passing.

lib/dry/configurable/instance_methods.rb
```diff
@@ -33,9 +33,7 @@ module InstanceMethods
       # Finalize the config and freeze the object
       #
       # @api public
-      def finalize!
-        return self if frozen?
```
ojab, line 37: `#frozen?` checks would be unreliable now and I don't think that it's performance-critical part, just drop them.

ojab:
Rebased, anything to be done to get it merged?

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -328,7 +328,7 @@
         klass.setting :env
 
         klass.setting :db do
-          setting :user, "root"
+          setting :user, +"root"
```
solnic, line 331: what...does this do?

ojab: /shrug [unfreezing the string](https://ruby-doc.org/core-2.7.2/String.html#method-i-2B)
Changed to `dup` to make it less confusing.

flash-gordon (approved):

lib/dry/configurable/instance_methods.rb
solnic: @ojab I wonder if this should just be idempotent or maybe if we should raise an error if a config is finalized and something tries to finalize it again 🤔

ojab:
Gentle ping.

lib/dry/configurable/instance_methods.rb
ojab: Oh my, missed this comment. We can raise an error, but now it's idempotent, so didn't changed it.
IMHO raising an error only would lead to people figuring how to workaround it and will not catch any real issues.

solnic (commented):
Would it be possible to just use `freeze` method and remove `finalize!`? Then `Config#freeze` could do its thing and behave just like `Hash#freeze` and THEN we could have `Config#deep_freeze` to make things simpler and more explicit.

This would be more flexible and possibly future-proof given that there's [a discussion about adding `Object#deep_freeze` to Ruby](https://bugs.ruby-lang.org/issues/17145).

I think getting this functionality done right is going to be crucial, that's why I'd prefer to really think it through. I should mention I'm integrating dry-configurable into rom-rb's core library and it's going to play major role there. So now all of a sudden I'm deeply interested in having proper config/settings/values freezing 😅 I used `finalize!` there and of course it didn't do what I assumed it would.

ojab:
>Then Config#freeze could do its thing and behave just like Hash#freeze 

umm, do you mean with recursion or without? Since it's not recursive/behaves like `Hash#freeze` right now and IMHO it's pretty useless, making it recursive would make it different from built-in objects. AFAIU it's `finalize!` name for that exact reason.
OTOH there is no discussion in the [PR](https://github.com/dry-rb/dry-configurable/pull/38) and it was 4 years ago + it's useless right now, so making it recursive would be sound.  

>and THEN we could have Config#deep_freeze to make things simpler and more explicit
>This would be more flexible and possibly future-proof given that there's a discussion about adding Object#deep_freeze to Ruby.

Heh, maybe with ractors `#deep_freeze` proposal would be more successful. But `#deep_freeze` is not in ruby yet, so there is non-zero chance to have different semantics and/or breaking change in dry-configurable when it will be.
Also `#deep_freeze` implies that we're freezing _everything_ recursively, which wouldn't be the case for the nested objects in setting values (e. g. for `Config.new(x: ['string']).deep_freeze` only array would be frozen and `'string'` wouldn't be), IMHO this would be unexpected.

And underlying implementation probably wouldn't change (two separate recursive freezers would be more complicated, I guess), so feel free to merge it as is and make `finalize!` private + bikeshed pulic names later :stuck_out_tongue_winking_eye:

AMHOL:
`#finalize!` was just a pattern that got copied over from `rom-rb` into `dry-container` and then here in `dry-configurable`, wasn't much thought that went into it other than that. I just thought of `#finalize!` as a method to be called at the end of the setup-phaze of an application, before any requests are made, and it made sense as a good place to freeze values.

ojab:
Oh, I'm not really accustomed to dry/rom ecosystem, with prior art it's perfectly logical naming without further justification. 
>made sense as a good place to freeze values.

part make me think that _maybe_ we should break API and make it freeze values by default, since we're not 1.0 yet ._. (dunno if dry ecosystem follows semver (google says https://github.com/dry-rb/dry-rb.org/issues/112#issuecomment-258213788, so API could be broken easily /shrug)).
I'd even made freezing opt-out in `.configure` and wherever else, but while dry-* is a bunch of separate gems, in reality it's quite tightly coupled (e. g. dry-validation pokes dry-schema `@private` apis) codebase that couldn't be changed independently, so it's probably wishful thinking.

ojab:
:crying_cat_face:

ojab:
:crying_cat_face:

solnic:
> 😿

We'll address this eventually. I'm just going back to OSS after a 3+ months break.

timriley (changes requested):
Apologies for the delay on our end, @ojab.

Overall I'm happy to see this go in. I think we should try and exercise this as much as we can between now and 1.0 to make sure we're happy with the API and behaviour. Left one small request for you, @ojab, if you don't mind :)

lib/dry/configurable/setting.rb
```diff
@@ -76,8 +76,11 @@ def input_defined?
 
       # @api private
       def value
-        @value ||= evaluate
+        return @value if evaluated?
+
+        @value = constructor[Undefined.coalesce(input, default, nil)]
       end
+      alias_method :evaluate, :value
```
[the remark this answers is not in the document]
ojab: Done (and rebased/fixed deprecation warnings).

timriley:
@solnic @flash-gordon Final check: we all good to merge this? I'm good with it, and given my other recent work here, I can roll this into a release next week.

timriley:
@ojab Just noticed you pushed some changes, and I should have checked with you too, sorry! Are you OK with the current state of this PR?

ojab:
Yeah, I just rebased since there was a merge conflict and I noticed it after opening/reading you comment in GH UI, no worries.

flash-gordon:
@timriley yeah I'm good with it though I don't remember all the details :)

solnic (approved):

ojab:
:crying_cat_face:

timriley:
@ojab I'll release this in the next few days!

timriley (approved):

timriley:
@ojab Thank you not only for your patch, but for your incredible patience and persistence here. I'd like to give you a trophy of some kind, but instead of hope you can settle for a new dry-configurable release 😄 Merging this one now, and will check your other PR (#131) next.

timriley:
Released in 0.15.0: https://github.com/dry-rb/dry-configurable/releases/tag/v0.15.0

PR #111, opened 2021-05-17
Deprecate constructor as a block

waiting-for-dev:
_Continues from #110_

---

So far, we have been accepting both a block and the `constructor:` keyword argument for providing a constructor. The former will not be available in 1.0, so we're deprecating it now.

We also prepare the test suite for the change and make naming consistent using "constructor" where we were using "pre-processor."

### DIFF
--- a/docsite/source/index.html.md
+++ b/docsite/source/index.html.md
@@ -28,7 +28,7 @@ class App
   end
   # Defaults to nil if no default value is given
   setting :adapter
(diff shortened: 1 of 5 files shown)


timriley (approved):
Looks good, happy for you to merge when you're ready!

waiting-for-dev:
Cool, thanks @timriley !

PR #112, opened 2021-05-17
Take default value from a keyword argument

waiting-for-dev:
From now on, the default value for a setting must be provided through
the `default:` keyword argument. I.e.:

```ruby
setting :name, default: 'Alice'
```

Providing it as a positional argument is still supported but deprecated.

@timriley this is basically your code on #86 (so feel free to add you as commit author :sweat_smile: )

@solnic , continuing from https://github.com/dry-rb/dry-configurable/pull/110#discussion_r632913069, extracting the deprecation expectations is a bit tricky because of global state + the need to rewind IO before asserting. So I just repeated the code for now. We could just remove the expectation altogether, as it's not that important and it'll be removed when version 1.0 is released.


timriley (approved):
Looks good to me, @waiting-for-dev!

If you wanted to put [co-author information](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) on the commit message, that might be fun, but I'm good either way really, just happy to see the work moving forward :)

I also think the duplication of test setup for the logger is fine, this is just for deprecation messages after all, and these tests should go away in the release where we remove the deprecations.

Happy for you to merge when you're ready. Looking forward to figuring out a release plan from here!

waiting-for-dev:
Thanks @timriley.

I added you as a co-author :slightly_smiling_face:

However, it seems that the tests are failing for ruby 3.0, I'll try to fix it tomorrow.

waiting-for-dev:
Fixed for 3.0 and merged :rocket:

PR #113, opened 2021-05-18
Forward keyword args to initialize

timriley:
This fixes an ArgumentError that would have been raised for classes including `Dry::Configurable` whose `#initialize` has required keyword parameters.

This is required to convert the base `Hanami::Configuration` class to dry-configurable, since that class has the following `#initialize` (whose params signature I'd like to retain):

```ruby
def initialize(env:)
```

### DIFF
--- a/lib/dry/configurable/instance_methods.rb
+++ b/lib/dry/configurable/instance_methods.rb
@@ -11,9 +11,16 @@ module Dry
     # @api private
     module Initializer
       # @api private
-      def initialize(*)
-        @config = Config.new(self.class._settings.dup)
-        super
+      if RUBY_VERSION >= "2.7"
+        def initialize(*, **)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
+      else
+        def initialize(*)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
       end
     end
 
--- a/lib/dry/configurable/instance_methods.rb
+++ b/lib/dry/configurable/instance_methods.rb
@@ -11,17 +11,11 @@ module Dry
     # @api private
     module Initializer
       # @api private
-      if RUBY_VERSION >= "2.7"
-        def initialize(*, **)
-          @config = Config.new(self.class._settings.dup)
-          super
-        end
-      else
-        def initialize(*)
-          @config = Config.new(self.class._settings.dup)
-          super
-        end
+      def initialize(*)
+        @config = Config.new(self.class._settings.dup)
+        super
       end
+      ruby2_keywords(:initialize) if respond_to?(:ruby2_keywords, true)
     end
 
     # Instance-level API when `Dry::Configurable` is included in a class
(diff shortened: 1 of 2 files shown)


lib/dry/configurable/instance_methods.rb
```diff
@@ -11,9 +11,16 @@ module Configurable
...
+        def initialize(*, **)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
+      else
+        def initialize(*)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
       end
```
timriley, line 24: To keep compatibility with 2.6, I tried various incantations of this method using `ruby2_keywords` (as suggested on https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/), but I couldn't get any to work.

Given this is a simple method, and given we're likely to drop support 2.6 fairly soon, and that we're not having to do this anywhere else within the lib, I'm OK with the duplication here.

Maybe you might have some ideas about how to handle this better, @flash-gordon?

timriley:
@solnic @flash-gordon lemme know if you're happy for this to go in :)

lib/dry/configurable/instance_methods.rb
flash-gordon: doesn't this work?
```ruby
    module Initializer
      # @api private
      def initialize(*)
        @config = Config.new(self.class._settings.dup)
        super
      end
      ruby2_keywords(:initialize) if respond_to?(:ruby2_keywords, true)
    end
```
I tested locally on 3.0, it passed

lib/dry/configurable/instance_methods.rb
```diff
@@ -11,9 +11,16 @@ module Configurable
...
+      if RUBY_VERSION >= "2.7"
+        def initialize(*, **)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
+      else
+        def initialize(*)
+          @config = Config.new(self.class._settings.dup)
+          super
+        end
```
solnic, line 23: I *think* we can levarage `...` in >= 3.0 too, so:

```ruby
if RUBY_VERSION >= "2.7"
  def initialize(...)
     @config = Config.new(self.class._settings.dup)
     super
  end
end
```

lib/dry/configurable/instance_methods.rb
timriley: @flash-gordon it didn't work with the older versions, IIRC

timriley: @flash-gordon By gum, you're right! I think I was trying it with `*args, **kwargs` and that was messing it up.

timriley:
@solnic @flash-gordon Pushed another commit showing the simplified version working. If you're good with this I'll do a squash when I merge the PR.

lib/dry/configurable/instance_methods.rb
flash-gordon: 1) it's 2.7
2) it'll require using strings and class_eval since it's a syntax change. I'd postpone `...` till 2.6 is EOL

flash-gordon (approved):

solnic (approved):

PR #115, opened 2021-06-03
Add the significant frame information to deprecation messages

waiting-for-dev:
It makes more easy to fix the deprecations introduced in #111 & #112

### DIFF
--- a/lib/dry/configurable/dsl.rb
+++ b/lib/dry/configurable/dsl.rb
@@ -41,7 +41,8 @@ module Dry
           Dry::Core::Deprecations.announce(
             "default value as positional argument to settings",
             "Provide a `default:` keyword argument instead",
-            tag: "dry-configurable"
+            tag: "dry-configurable",
+            uplevel: 2
           )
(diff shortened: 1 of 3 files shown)


solnic (approved):

flash-gordon:
For the record, it didn't help much in this case:
```
dry-effects/bundler/gems/dry-configurable-00f8b8e64898/lib/dry/configurable/class_methods.rb:41:in `setting' [dry-configurable] default value as positional argument to settings is deprecated and will be removed in the next major version
```
It points to this line https://github.com/dry-rb/dry-configurable/blob/master/lib/dry/configurable/class_methods.rb#L41 which is in the gem itself :)

waiting-for-dev:
Hmmm... I made sure that it went beyond the gem. In fact, it's something that is tested. Not sure how it can behave differently in other places. How did you call the `setting` method?

timriley:
I saw this issue too, and I thought it might need a fix in dry-core, hence https://github.com/dry-rb/dry-core/pull/56

PR #117, opened 2021-07-04
Add convenience method Dry::Configurable::Settings#replace

### LINKED ISSUE #109: Add `Settings#replace` to support wholesale replacement of settings in a configurable object
This would better support the use case in https://github.com/dry-rb/dry-system/pull/162 (see https://github.com/dry-rb/dry-system/pull/162#discussion_r613203242 for specific notes), in which all the settings from one class are "imported" into another for use there too.

Currently, this is done like so:

```ruby
AnotherConfigurableThing._settings.each do |setting|
  _settings << setting.dup
end
```

But it would be nicer to do it like this:

```ruby
_settings.replace(AnotherConfigurableThing._settings.dup)
```

skinnyjames:
Hiya,
Really new to this project and dry-rb in general, so I don't expect to hit the mark on my first try, but this PR fixes #109 

I'm curious if this is the desired api, or if it would be easier to copy settings from another Dry::Configurable class or object without accessing what I am assuming to be a an internal var?

I was thinking..

```ruby
  class One
    extend Dry::Configurable
    setting :hello, "world"
  end

  class Two
    extend Dry::Configurable
    setting :hello, "bazz"
  end

  Two.clobber_config(One)
  
  One.config.hello #=> "World"
  Two.config.hello #=> "World"
```

### DIFF
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -44,7 +44,7 @@ module Dry
 
         def merge(setting)
           settings = input.merge(setting.input)
-          Nested.new(name, input: settings, **options)
+          Nested.new(name, input: settings, **setting.options)
         end
 
         # @api private
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -69,6 +69,11 @@ module Dry
         evaluate if input_defined?
       end
 
+      def merge(setting)
+        input.merge(setting.input) if input_defined?
+        evaluate
+      end
+
       # @api private
       def input_defined?
         !input.equal?(Undefined)
--- a/lib/dry/configurable/setting.rb
+++ b/lib/dry/configurable/setting.rb
@@ -42,6 +42,11 @@ module Dry
       class Nested < Setting
         CONSTRUCTOR = Config.method(:new)
 
+        def merge(setting)
+          settings = input.merge(setting.input)
+          Nested.new(name, input: settings, **options)
+        end
+
         # @api private
         def pristine
           with(input: input.pristine)
@@ -67,11 +72,6 @@ module Dry
         @options = options
 
         evaluate if input_defined?
-      end
-
-      def merge(setting)
-        input.merge(setting.input) if input_defined?
-        evaluate
       end
 
       # @api private
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -71,7 +71,7 @@ module Dry
       private
 
       def merge_setting(setting)
-        if elements[setting.name].is_a? Setting::Nested
+        if elements[setting.name].is_a?(Setting::Nested) && setting.is_a?(Setting::Nested)
           elements[setting.name].merge(setting)
         else
           setting
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -34,11 +34,15 @@ module Dry
       end
 
       def merge(settings)
+        unless settings.is_a? Dry::Configurable::Settings
+          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
+        end
         settings.each do |setting|
           merge_setting(setting)
         end
       end
 
+      # @api private
       def merge_setting(setting)
         if key?(setting.name)
           elements[setting.name].merge(setting)
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -27,7 +27,7 @@ module Dry
         unless settings.is_a? Dry::Configurable::Settings
           raise ArgumentError, "settings must be a Dry::Configurable::Settings"
         end
-        
+
         settings.each do |setting|
           self << setting
         end
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -37,7 +37,7 @@ module Dry
         unless settings.is_a? Dry::Configurable::Settings
           raise ArgumentError, "settings must be a Dry::Configurable::Settings"
         end
-        
+
         settings.each do |setting|
           merge_setting(setting)
         end
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -23,32 +23,17 @@ module Dry
         initialize_elements(elements)
       end
 
-      def replace(settings)
-        unless settings.is_a? Dry::Configurable::Settings
-          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
-        end
-
-        settings.each do |setting|
+      def merge!(settings)
+        merge(settings).each do |setting|
           self << setting
         end
       end
 
       def merge(settings)
-        unless settings.is_a? Dry::Configurable::Settings
-          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
-        end
+        ensure_arguments(settings)
 
-        settings.each do |setting|
-          merge_setting(setting)
-        end
-      end
-
-      # @api private
-      def merge_setting(setting)
-        if key?(setting.name)
-          elements[setting.name].merge(setting)
-        else
-          self << setting
+        settings.dup.inject(dup) do |memo, setting|
+          memo << merge_setting(setting)
         end
       end
 
@@ -84,6 +69,20 @@ module Dry
       end
 
       private
+
+      def merge_setting(setting)
+        if elements[setting.name].is_a? Setting::Nested
+          elements[setting.name].merge(setting)
+        else
+          setting
+        end
+      end
+
+      def ensure_arguments(settings)
+        unless settings.is_a? Dry::Configurable::Settings
+          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
+        end
+      end
 
       # @api private
       def initialize_copy(source)
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -37,6 +37,7 @@ module Dry
         unless settings.is_a? Dry::Configurable::Settings
           raise ArgumentError, "settings must be a Dry::Configurable::Settings"
         end
+        
         settings.each do |setting|
           merge_setting(setting)
         end
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -33,6 +33,20 @@ module Dry
         end
       end
 
+      def merge(settings)
+        settings.each do |setting|
+          merge_setting(setting)
+        end
+      end
+
+      def merge_setting(setting)
+        if key?(setting.name)
+          elements[setting.name].merge(setting)
+        else
+          self << setting
+        end
+      end
+
       # @api private
       def <<(setting)
         elements[setting.name] = setting
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -24,7 +24,10 @@ module Dry
       end
 
       def replace(settings)
-        raise ArgumentError, "settings must be a Dry::Configurable::Settings" unless settings.is_a? Dry::Configurable::Settings
+        unless settings.is_a? Dry::Configurable::Settings
+          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
+        end
+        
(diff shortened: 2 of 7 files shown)


solnic:
Thanks for working on this. It would be great to expose `Settings` object via `#settings` method, rather than this set object which is just some legacy we have (@timriley do you remember why it's even there?). Then you could just do things like `ChildClass.settings.replace(ParentClass.settings)` etc.

I'm working on integrating dry-configurable into rom-rb's core at the moment and methods like `replace` or `merge` or `append` etc. would be *incredibly useful*. I actually think some "settings and config algebra" should be a 1st-class feature in this library and it would make it really stand out. There are a lot of interesting usage patterns that involve merging/inheriting configuration that can actually be really tricky to deal with manually.

skinnyjames:
No problem!
I didn't realize that I was still failing checks.  I'll get that cleaned up tonight.
I can also work on merge / append too :)

flash-gordon:
@skinnyjames sure, `bundle exec rubocop` should do

skinnyjames:
Fixed up the linting hopefully!
I also added `Settings#merge` functionality in my origin branch 
 https://github.com/skinnyjames-forks/dry-configurable/pull/1/files
 
 If that seems like a good add, I'm happy to add it to the PR.

timriley:
> It would be great to expose `Settings` object via `#settings` method, rather than this set object which is just some legacy we have (@timriley do you remember why it's even there?). Then you could just do things like `ChildClass.settings.replace(ParentClass.settings)` etc.

@solnic I don't have any clue why it's the way it is, unfortunately! I presume we've simply been preserving some very early API decision without giving it a whole ton of thought.

I agree that the set object would be better replaced by the full `Settings` instance — let's consider that for a follow up PR? Be good to do that pre-1.0.

I'll leave some other thoughts on this PR itself in another comment shortly 😄

lib/dry/configurable/settings.rb
```diff
@@ -23,6 +23,35 @@ def initialize(elements = EMPTY_ARRAY)
...
+        unless settings.is_a? Dry::Configurable::Settings
+          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
+        end
+
+        settings.each do |setting|
+          self << setting
+        end
+      end
+
+      def merge(settings)
```
timriley, line 36: Before we commit to this naming, I think it'd be worth contemplating whether we want a `#merge` that returns a new copy of the `Settings`, and a `merge!` that modifies in place (which this method as implemented is currently doing).

What do you think, @solnic? If we're going to follow that idea of yours and make the `Settings` more directly accessible via a `.settings` class method (instead of `._settings`), then having a slightly more expansive API on those objects could be handy.

(`#replace` OTOH doesn't need the bang on the end of its name because its only function is to mutate the receiver - we're mirroring the Ruby stdlib in this regard)

lib/dry/configurable/settings.rb
```diff
@@ -23,6 +23,35 @@ def initialize(elements = EMPTY_ARRAY)
...
+
+        settings.each do |setting|
+          merge_setting(setting)
+        end
+      end
+
+      # @api private
+      def merge_setting(setting)
+        if key?(setting.name)
+          elements[setting.name].merge(setting)
```
timriley, line 49: This branch feels like it might need some more expansive handling. Right now if we had a settings object containing `a`, `b`, and `c` settings, and they're all primitive values, e.g. strings, and we merged in a settings object with just an `a` string, then what I would expect is the `a` from the second settings to _replace_ the first, which won't happen given the current code, AFAICT.

I think we should expand the test examples to encompass cases like these and adjust the implementation to suit.

One way to handle this could be to check whether `elements[setting.name]` is a `Setting::Nested` and then handle either case appropriately.

lib/dry/configurable/settings.rb
```diff
@@ -23,6 +23,35 @@ def initialize(elements = EMPTY_ARRAY)
...
+          raise ArgumentError, "settings must be a Dry::Configurable::Settings"
+        end
+
+        settings.each do |setting|
+          merge_setting(setting)
+        end
+      end
+
+      # @api private
+      def merge_setting(setting)
```
timriley, line 47: BTW, this method should be moved below the `private` line (and we can then ditch the `# @api private` comment at the same time).

lib/dry/configurable/setting.rb
```diff
@@ -69,6 +69,11 @@ def initialize(name, input: Undefined, default: Undefined, **options)
         evaluate if input_defined?
       end
 
+      def merge(setting)
+        input.merge(setting.input) if input_defined?
+        evaluate
+      end
```
timriley, line 75: Depending on how we adjust `Settings#merge_setting` (per my previous comment), I expect we might want to make some changes here too.

For example, it probably makes sense to move this into `Setting::Nested`, because merging a non-nested setting doesn't really make sense.

At the same time, we should probably raise an `ArgumentError` if the given `setting` param is not also an instance of `Setting::Nested`.

And once we're inside `Setting::Nested`, we could probably make the implementation a lot more direct. The `input_defined?` check won't be needed because `Nested` settings will _always_ have an input. And then because we're no longer having to worry about constructors or default values (not relevant for nested settings), we could probably skip the indirection of `evaluate` and just set the `@value` directly.

You'll have to give it a try to see how these changes shake out, but I think it'd be worth it!

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -185,6 +185,68 @@
 
     include_context "configurable behavior"
 
+    context "can be configured with another class's settings" do
```
timriley, line 188: FWIW, this file is actually testing the `.setting` method (n.b. the `RSpec.describe ... ".setting"` call at the top of the file), and these new examples are probably outside of that scope. It'd probably be helpful to move them into a new `spec/.integration/.../settings_spec.rb` file.

timriley (changes requested):
Thanks for getting to work on this, @skinnyjames! Really appreciate your effort!

Now, I hope you don't mind, but I've left a few somewhat extensive comments on the code here. We're definitely heading in the right direction, but there are some details we'll want to get right and extra scenarios we should consider before this is a fully resolved new feature.

Anyway, please have a read and let me know if you think anything needs clarifying. Am definitely happy to work with you (and @solnic) as we iterate on this feature!

timriley:
BTW, @skinnyjames, regarding your original comments in the PR description:

> I'm curious if this is the desired api, or if it would be easier to copy settings from another Dry::Configurable class or object without accessing what I am assuming to be a an internal var?
> 
> I was thinking..
> 
> ```ruby
>   class One
>     extend Dry::Configurable
>     setting :hello, "world"
>   end
> 
>   class Two
>     extend Dry::Configurable
>     setting :hello, "bazz"
>   end
> 
>   Two.clobber_config(One)
>   
>   One.config.hello #=> "World"
>   Two.config.hello #=> "World"
> ```

I think the approach you chose in the actual code changes you've made is the right one. 👍🏼

Since `Dry::Configurable` can be mixed into all sorts of classes, I don't think we want to add too many extra methods to those classes. By sticking to `setting` defining new settings, and then `settings` (`_settings` right now, but I think we can rename it, per my comments to @solnic earlier), we can then allow most of the extra behaviour to hang off that `settings` object rather than polluting the method list of whatever class is mixing in our module.

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -185,6 +185,68 @@
...
+          end
+        end
+
+        other_klass.setting :database do
+          setting :host, "localhost"
+          setting :port
+          setting :deep, 3
+        end
+
+        other_klass._settings.merge(klass._settings.dup)
```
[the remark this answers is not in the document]
skinnyjames: > Thanks for getting to work on this, @skinnyjames! Really appreciate your effort!
> 
> Now, I hope you don't mind, but I've left a few somewhat extensive comments on the code here. We're definitely heading in the right direction, but there are some details we'll want to get right and extra scenarios we should consider before this is a fully resolved new feature.
> 
> Anyway, please have a read and let me know if you think anything needs clarifying. Am definitely happy to work with you (and @solnic) as we iterate on this feature!

Hey thanks for the feedback! 
Definitely trying to improve, so the detailed feedback is great.
Everything you've said makes sense, and I'll work on addressing them tonight!

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -185,6 +185,68 @@
...
+        expect(other_klass.config.database.deep).to eql(3)
+      end
+
+      it "replaces with each" do
+        klass.setting :hello, "world"
+        klass._settings.each do |setting|
+          other_klass._settings << setting.dup
+        end
+        expect(other_klass.config.hello).to eql("world")
+      end
```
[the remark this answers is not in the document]
skinnyjames: I was trying to lock down the expected usage so that I didn't break anything, but it seems unneeded.

lib/dry/configurable/settings.rb
skinnyjames: @timriley I'm thinking about this, and got a little tripped up.

Typically merge does replace the fields on the first, like you said.  I'm not exactly sure what I was thinking.
But I'm also having trouble distinguishing the difference between a merge and a deep replace.
Should replace delete fields that don't exist on the settings object that is being replaced?

Also I feel like merge tends to need to be opinionated on merge strategies.
For instance, how would you want to handle the merging of array / hash values? (or should `#merge` replace any value that isn't a Setting?)

lib/dry/configurable/settings.rb
skinnyjames: I'm unsure of the difference between a merge! and replace, so I removed replaced in favor of merge, and merge!

lib/dry/configurable/settings.rb
skinnyjames: The newest commit should replace any nested value that isn't a Settings::Nested.
There is no handling currently for merging primitives.

skinnyjames:
I'm noticing a weird state issue when invoking the config accessor before merging the settings.
this passes.
```ruby
        klass.setting :database do
          setting :dsn, "localhost"
        end
        other_klass.setting :database do
          setting :dsn, "remote"
        end

        other_klass._settings.merge!(klass._settings)
        expect(other_klass.config.database.dsn).to eql("localhost")
```

but this fails.
```ruby
        klass.setting :database do
          setting :dsn, "localhost"
        end
        other_klass.setting :database do
          setting :dsn, "remote"
        end
        expect(other_klass.config.database.dsn).to eql("remote")
        other_klass._settings.merge!(klass._settings)
        expect(other_klass.config.database.dsn).to eql("localhost")
```

For now, I'm not at all sure why that is happening.
adding failing tests in the meantime.

lib/dry/configurable/settings.rb
timriley: I think this is definitely the right call 👍🏼

lib/dry/configurable/settings.rb
timriley: I think we had a bit of confusion just then. I wasn't advocating for us to rename `#replace` – that's a good name and we should keep it as it was before. Would you mind putting that back?

What I was actually musing on was about the behaviour of the `Settings#merge` method, whether it should mutate the `Settings` instance or return a _new copy_ of the the instance with the merged settings, leaving the original one alone. I think it might make sense of us to have both. But for the time being, let's rename your original `#merge` to `#merge!` for clarity.

timriley:
Thanks @skinnyjames — just left you a couple more notes in the previous batch of comments. I'll take a look at the failing specs in the next day or two, I hope.

lib/dry/configurable/settings.rb
skinnyjames: I can add replace back no problem, but the PR fix I made also already adds a non destructive "merge" and a destructive "merge!", like you said.

I'll go ahead and add replace back.

skinnyjames:
> Thanks @skinnyjames — just left you a couple more notes in the previous batch of comments. I'll take a look at the failing specs in the next day or two, I hope.

No problem.  The api should be a little more clear now that "merge" returns settings, and "merge!" will merge those settings.  I think the reason the tests will fail is because the config is cached once invoked.  It seems intentional though.

Issue #120, opened 2021-08-05
0.13.0 release checklist


timriley:
The upcoming 0.13.0 release will include a couple of large changes:

1. Deprecating setting constructors provided via a block, favouring the `constructor:` argument instead
2. Requiring default values be provided as a `default:` argument, rather than a second positional argument

While we've strived for these changes to be backwards compatible, they may cause breakage in certain circumstances.

To this end, we need to make sure we coordinate this 0.13.0 release of dry-configurable with all the gems that use it within the dry-rb, rom-rb, and Hanami ecosystems. This issue will serve as this coordination point.

## Testing dependent gems

To make sure the new dry-configurable 0.13.0 release offers the appropriate level of backwards compatibility, we want to test all of its dependent gems across the dry-rb/rom-rb/hanami ecosystem.

The table below tracks the test we're (manually) running on each gem. The tests should exercise each of these cases:

- The latest release of the gem, using dry-configurable 0.12.1, on Ruby 3.0
- The latest release of the gem, using dry-configurable 0.12.1, on Ruby 2.7
- The latest release of the gem, using dry-configurable 0.13.0 (master), on Ruby 3.0
- The latest release of the gem, using dry-configurable 0.13.0 (master), on Ruby 2.7
- The branch of the gem with dry-configurable 0.13.0 updates, using dry-configurable 0.13.0 (master), on Ruby 3.0
- The branch of the gem with dry-configurable 0.13.0 updates, using dry-configurable 0.13.0 (master), on Ruby 2.7

The tests should pass in all cases.

Testing instructions are as follows:

1. In your working copy, run `git fetch --all --tags` to get all tags
2. Check out the the branch or tag for the test, e.g. `git checkout tags/v0.8.0` or `git checkout master`
3. Ensure the gemspec points to the appropriate version of dry-configurable for the test, e.g. `0.12.1` or the `master` branch
4. Run the test suite and ensure it passes

| Dependent gem | dry-configurable 0.12.1 | dry-configurable 0.13.0 |
| -- | -- | -- |
| dry-container @ v0.8.0 @ ruby 3.0 | ✅ | ✅ |
| dry-container @ v0.8.0 @ ruby 2.7 | ✅ | ✅ |
| dry-container @ master @ ruby 3.0 | n/a | ✅ |
| dry-container @ master @ ruby 2.7 | n/a | ✅ |
[...]

PR #121, opened 2021-08-05
Properly support and deprecate default hash values provided as second positional argument to `setting` for Ruby 2.6 and 2.7

timriley:
Before this change, if we defined a setting like this:

```ruby
setting :default_options, {foo: "bar"}
```

It would fail on Ruby 2.6 and 2.7 because inside `DSL#setting`, whose signature is this:

```
def setting(name, default = Undefined, **options, &block)
```

The `default` value remains as `Undefined`, and `options` becomes `{foo: "bar"}`, which then sees an exception raised inside `#ensure_valid_options` because (of course) they are not valid options for `setting`.

What we really want is for that hash provided as the second positional parameter to be assigned to `default`, and `options` to remain as `{}`.

---

Per the extended code comments included in the change, we take the following approach to handle this case:

In Ruby 2.6 and 2.7, when a hash is given as the second positional argument (i.e. the hash is intended to be the setting's default value), and there are no other keyword arguments given, the hash is assigned to the `options` variable instead of `default`.

For example, for this setting:

```ruby
setting :hash_setting, {my_hash: true}
```

We'll have a `default` of `Undefined` and an `options` of `{my_hash: true}`

If any additional keyword arguments are provided, e.g.:

```ruby
setting :hash_setting, {my_hash: true}, reader: true
```

Then we'll have a `default` of `{my_hash: true}` and an `options` of `{reader: true}`, which is what we want.

To work around that first case and ensure our (deprecated) backwards compatibility holds for Ruby 2.6 and 2.7, we extract all invalid options from `options`, and if there are no remaining valid options (i.e. if there were no keyword arguments given), then we can infer the invalid options to be a default hash value for the setting.

This approach also preserves the behavior of raising an ArgumentError when a distinct hash is _not_ intentionally provided as the second positional argument (i.e. it's not enclosed in braces), and instead invalid keyword arguments are given alongside valid ones. So this setting:

```ruby
setting :some_setting, invalid_option: true, reader: true
```

Would raise an ArgumentError as expected.

However, the one case we can't catch here is when invalid options are supplied without hash literal braces, but there are no other keyword arguments supplied. In this case, a setting like:

```ruby
setting :hash_setting, my_hash: true
```

Is parsed identically to the first case described above:

```ruby
setting :hash_setting, {my_hash: true}
```

So in both of these cases, the default value will become `{my_hash: true}`. We consider this unlikely to be a problem in practice, since users are not likely to be providing invalid options to `setting` and expecting them to be ignored. Additionally, the deprecation messages will make the new behavior obvious, and encourage the users to upgrade their setting definitions.

### DIFF
--- a/lib/dry/configurable/dsl.rb
+++ b/lib/dry/configurable/dsl.rb
@@ -27,10 +27,10 @@ module Dry
         instance_exec(&block) if block
       end
 
-      # Register a new setting node and compile it into a setting object
+      # Registers a new setting node and compile it into a setting object
       #
       # @see ClassMethods.setting
-      # @api public
+      # @api private
       # @return Setting
       def setting(name, default = Undefined, **options, &block)
         unless VALID_NAME.match?(name.to_s)
--- a/lib/dry/configurable/dsl.rb
+++ b/lib/dry/configurable/dsl.rb
@@ -32,7 +32,7 @@ module Dry
       # @see ClassMethods.setting
       # @api private
       # @return Setting
-      def setting(name, default = Undefined, **options, &block)
+      def setting(name, default = Undefined, **options, &block) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
         unless VALID_NAME.match?(name.to_s)
           raise ArgumentError, "#{name} is not a valid setting name"
         end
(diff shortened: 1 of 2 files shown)


timriley:
@flash-gordon @solnic are you aware of any trickery that might let us get these arguments into the right local variables?

timriley:
A possible hack (but potentially good enough given this is a backwards compatibility shim and it's only problematic on older Rubies): if `default` is `Undefined`, and `options` is a non-empty hash, then take all the valid keys/vals out of `options`, and if there is still a non-empty hash remaining, treat _that_ as the `default`. We could also add some extra deprecation warnings for the special case, too, e.g. "We're inferring a default value from your provided hash, please provide it as `default:` instead blah blah".

One thing I should note here is that this is _only_ a problem if there are no other options provided, so this case below is actually fine:

```ruby
setting :foo, {foo: "bar"}, reader: true
```

In this case, the `default` is `{foo: "bar"`} and options is `reader: true`.

flash-gordon:
> A possible hack (but potentially good enough given this is a backwards compatibility shim and it's only problematic on older Rubies?): if default is Undefined, and options is a non-empty hash, then take all the valid keys out of options, and if there is still a non-empty hash remaining, treat that as the default. We could also add some extra deprecation warnings for the special case, too, e.g. "We're inferring a default value from your provided hash, please provide it as default: instead blah blah".

I'd go with this. I don't think bothering with warnings is worth it though

timriley:
@flash-gordon Thanks for the feedback :) Yeah, it felt reasonable to me too when I thought of it — I'll go ahead and prepare this, then!

AMHOL:
Had this problem a while back https://github.com/dry-rb/dry-configurable/pull/14

Your proposed solution sounds much nicer than mine :D

timriley:
@solnic @flash-gordon I've addressed this issue now and updated the PR description to describe my approach.

Would you mind checking this out and letting me know if you're happy with it?

Once we merge this in, then it means the change for the default value from the second positional argument to the `default:` keyword argument is as backwards compatible as can possibly be, which should give us confidence to make this next dry-configurable release without having to employ any additional compatibility shims in the dry-rb gems that are already using it (i.e. following the approach I described in https://github.com/dry-rb/dry-configurable/issues/120#issuecomment-893449117).

I'll still do another couple of full rounds of testing inside the gems listed in https://github.com/dry-rb/dry-configurable/issues/120, of course, but I hope that should be quite straightforward after this adjustment here.

waiting-for-dev:
Thanks, @timriley, amazing work!

I agree with the solution. When we have both an unknown key and a known key, we have to favor one of the two possible options:

1 - The user wants to use a known key on its default hash.
2 - The user misspelled a known key on its kwargs.

At this point, known keys are:

https://github.com/dry-rb/dry-configurable/blob/793a2035187fab5db883fdd4d4de17884789b4ab/lib/dry/configurable/setting.rb#L18

So, probably, there're no big chances of name collision.

With your solution, we're going with the second option, which is better because users will get an error because of unknown keyword arguments, which will help them work around it.

Other than that, I think you'll need to remove those `FileUtils.pwd` before merging. Please, see #119 .

lib/dry/configurable/dsl.rb
```diff
@@ -77,8 +141,17 @@ def ensure_valid_options(options)
...
+
         raise ArgumentError, "Invalid options: #{invalid_keys.inspect}" unless invalid_keys.empty?
       end
+
+      # Returns a tuple of valid and invalid options hashes derived from the options hash
+      # given to the setting
+      def valid_and_invalid_options(options)
+        options.each_with_object([{}, {}]) do |(key, val), (valid_options, invalid_options)|
+          (Setting::OPTIONS.include?(key) ? valid_options : invalid_options)[key] = val
+        end
```
flash-gordon, line 153:
```ruby
options.partition { |k, _| Setting::OPTIONS.include?(k) }.map(&:to_h)
```

flash-gordon (approved):
LGTM 👍

lib/dry/configurable/dsl.rb
timriley: So much neater, thanks @flash-gordon! I've applied this now.

timriley:
Thanks for the feedback, folks! Looks like we're all in agreement on approach here. I'm going to merge this now so I can get on with the testing.

PR #130, opened 2022-01-07
Allow settings to be defined and accessible after class-level `config` access

timriley:
With this change, it's now possible to access the class-level `config` object and still have subsequently defined settings still available on the `config`. For example:

```ruby
class MyClass
  extend Dry::Configurable

  setting :foo, default: "foo"
  config
  setting :bar, default: "bar"
end

MyClass.config._settings.map(&:name) # => [:foo, :bar]

MyClass.config.bar = "new bar"
MyClass.config.bar # => "new bar"
```

Before this change, the `config._settings` list would have been just `[:foo]` and you would have received a `NoMethodError` when trying to call `config.bar=`.

This change works by having `Config` no longer eagerly `dup` the `settings` it receives when it initializes. This allows it to share the same object as the class-level `_settings` that receives new settings as they're defined.

We still ensure configurable subclasses have distinct settings and config by `dup`ing those objects at the time of subclassing.

We also ensure that configurable instances have distinct config by `dup`ing the settings at the time of initialization.

Originally, there were two incidental behavioral changes also introduced with this PR. The first _was_ that default values for settings were no longer eagerly duped when assigned, but with https://github.com/dry-rb/dry-configurable/pull/130/commits/a68c5f11e5263203103818a8860a3af66ee8801c I was able to preserve the existing behavior.

This leaves just the one incidental behavior change, demonstrated in the new `"configuring the parent after subclassing does not copy the config to the child"` spec example:

```ruby
specify "configuring the parent after subclassing does not copy the config to the child" do
  klass.setting :db

  subclass = Class.new(klass)

  object.config.db = "mariadb"

  expect(subclass.config.db).to be nil
end
```

In other words, if you:

1. Create a configurable parent class
2. Create a subclass
3. Change a config on that parent class
4. Then it will _not_ propagate to the child class

IMO, this actually feels like the right behavior versus what we had before, so I'm comfortable introducing it.

It's also more consistent with another closely-related existing test case, `"adding parent setting does not affect child"`, which follows the same pattern: create a configurable parent class, subclass it, do something to the parent class, do not expect it to flow through to the subclass.

---

The overall driver for this change is that it allows us to remove [this overloaded `setting` method](https://github.com/dry-rb/dry-system/blob/b7de4a9b4d9cb7bc2138a02847d6552a13eb2207/lib/dry/system/container.rb#L103-L113) from `Dry::System::Container`, an important in-ecosystem user of dry-configurable. See https://github.com/dry-rb/dry-system/pull/201 for this change in action.

Having dry-configurable serve this need out of the box for dry-system is a sign that this is a helpful change to be making for other users too, especially now while we're still pre-1.0 for dry-configurable.

### DIFF
--- a/spec/integration/dry/configurable/setting_spec.rb
+++ b/spec/integration/dry/configurable/setting_spec.rb
@@ -45,31 +45,12 @@ RSpec.describe Dry::Configurable, ".setting" do
     end
 
     context "with a default value" do
-      context "string" do
-        before do
-          klass.setting :db, default: "sqlite"
-        end
-
-        it "presets the default value" do
-          expect(object.config.db).to eql("sqlite")
-        end
+      before do
+        klass.setting :db, default: "sqlite"
       end
 
-      context "hash" do
-        it "returns the default value" do
-          klass.setting :db_config, default: {user: "root", password: ""}
-
-          expect(object.config.db_config).to eql(user: "root", password: "")
-        end
-
-        it "copies the original hash object" do
-          hash = {user: "root", password: ""}
-
-          klass.setting :db_config, default: hash
-
-          expect(object.config.db_config).to_not be(hash)
-          expect(object.config.db_config).to eql(hash)
-        end
+      it "presets the value with the default" do
+        expect(object.config.db).to eql("sqlite")
       end
     end
 
@@ -185,6 +166,23 @@ RSpec.describe Dry::Configurable, ".setting" do
 
     include_context "configurable behavior"
 
+    context "class-level configurable behavior" do
+      specify "settings defined after accessing config are still available in the config" do
+        klass.setting :before, default: "defined before"
+        klass.config
+        klass.setting :after, default: "defined after"
+
+        expect(klass.config.before).to eq "defined before"
+        expect(klass.config.after).to eq "defined after"
+      end
+
+      specify "default values use their original objects" do
+        hash = {user: "root", password: ""}
+        klass.setting :db_config, default: hash
+        expect(object.config.db_config).to be(hash)
+      end
+    end
+
     context "with a subclass" do
       let(:subclass) do
         Class.new(klass)
@@ -225,12 +223,22 @@ RSpec.describe Dry::Configurable, ".setting" do
         expect(subclass.settings).to eql(Set[:db])
       end
 
-      it "configured parent copies config to the child" do
+      specify "configuring the parent before subclassing copies the config to the child" do
         klass.setting :db
 
         object.config.db = "mariadb"
 
         expect(subclass.config.db).to eql("mariadb")
+      end
+
+      specify "configuring the parent after subclassing does not copy the config to the child" do
+        klass.setting :db
+
+        subclass = Class.new(klass)
+
+        object.config.db = "mariadb"
+
+        expect(subclass.config.db).to be nil
       end
 
       it "not configured parent does not set child config" do
@@ -321,6 +329,22 @@ RSpec.describe Dry::Configurable, ".setting" do
       expect(object.config.path).to eq Pathname("test")
       expect(new_object.config.path).to eq Pathname("test")
       expect(object.config.path).not_to be(new_object.config.path)
+    end
+
+    it "makes only settings defined before instantiation available" do
+      klass.setting :before, default: "defined before"
+
+      object_1 = klass.new
+
+      klass.setting :after, default: "defined after"
+
+      object_2 = klass.new
+
+      expect(object_1.config.before).to eq "defined before"
+      expect(object_1.config).not_to respond_to(:after)
+
+      expect(object_2.config.before).to eq "defined before"
+      expect(object_2.config.after).to eq "defined after"
     end
 
     shared_examples "copying" do
--- a/spec/integration/dry/configurable/setting_spec.rb
+++ b/spec/integration/dry/configurable/setting_spec.rb
@@ -45,12 +45,31 @@ RSpec.describe Dry::Configurable, ".setting" do
     end
 
     context "with a default value" do
-      before do
-        klass.setting :db, default: "sqlite"
+      context "string" do
+        before do
+          klass.setting :db, default: "sqlite"
+        end
+
+        it "presets the default value" do
+          expect(object.config.db).to eql("sqlite")
+        end
       end
 
-      it "presets the value with the default" do
-        expect(object.config.db).to eql("sqlite")
+      context "hash" do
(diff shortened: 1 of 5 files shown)


spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -323,6 +331,22 @@
...
+
+      object_2 = klass.new
+
+      expect(object_1.config.before).to eq "defined before"
+      expect(object_1.config).not_to respond_to(:after)
+
+      expect(object_2.config.before).to eq "defined before"
+      expect(object_2.config.after).to eq "defined after"
+    end
+
```
timriley, line 349: n.b. **This is not new behaviour** for this PR. I've copied this example into the master branch, and it passes there also. However, I thought it was useful to specify this behaviour clearly given we have slightly different behaviour at the class-level now.

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -185,6 +185,15 @@
...
     include_context "configurable behavior"
 
+    specify "settings defined after accessing config are still available in the config" do
+      klass.setting :before, default: "defined before"
+      klass.config
+      klass.setting :after, default: "defined after"
+
+      expect(klass.config.before).to eq "defined before"
+      expect(klass.config.after).to eq "defined after"
+    end
```
timriley, line 195: This is test case that exercises the new behavior I introduced with this change.

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -225,14 +234,24 @@
...
 
+      specify "configuring the parent after subclassing does not copy the config to the child" do
+        klass.setting :db
+
+        subclass = Class.new(klass)
+
+        object.config.db = "mariadb"
+
+        expect(subclass.config.db).to be nil
+      end
```
timriley, line 253: As mentioned in the PR description, this is the only other behavior change introduced here, but I think it's a reasonable one.

spec/integration/dry/configurable/setting_spec.rb
AMHOL: Perhaps it should raise in this case?

solnic (approved):

flash-gordon:
> IMO, this actually feels like the right behavior versus what we had before, so I'm comfortable introducing it.
>
> It's also more consistent with another closely-related existing test case, "adding parent setting does not affect child", which follows the same pattern: create a configurable parent class, subclass it, do something to the parent class, do not expect it to flow through to the subclass.

I kinda agree with that but this may clash with the idea of plugin-provided settings where new options are added at later stages. It'd be good to come up with an approach that works for this case, mb having an option for `.setting`. WDYT?

timriley:
> > IMO, this actually feels like the right behavior versus what we had before, so I'm comfortable introducing it.
> > It's also more consistent with another closely-related existing test case, "adding parent setting does not affect child", which follows the same pattern: create a configurable parent class, subclass it, do something to the parent class, do not expect it to flow through to the subclass.
> 
> I kinda agree with that but this may clash with the idea of plugin-provided settings where new options are added at later stages. It'd be good to come up with an approach that works for this case, mb having an option for `.setting`. WDYT?

@flash-gordon Yeah, this is a good case to consider.

I could see it going in a number of ways:

- We make the current (as of this PR) behavior clearly documented, and make it clear that our expectation for plugin-based systems that extend settings to ensure all plugins are loaded before configurable parent classes are subclassed. Given that it's typical for systems to initialize their plugins before loading user-provided code, this may be just fine.
- We make this behaviour configurable, like you suggested, at the expense of the added complexity.

I might take a quick spike at the configurability now, just to see what it might look like.

timriley:
> I might take a quick spike at the configurability now, just to see what it might look like.

I gave it a go, and honestly, I don't think there's an easy way to do this, at least not without overhauling the library again.

We treat settings and config as _state_, and at the point of class inheritance, there's a clear moment of _state transfer_, and to make sure that e.g. a child class doesn't end up idirectly configuring its parent, we duplicate that state at that moment of transfer.

This principle is also apparent in the related spec example that I mentioned above, too: `"adding parent setting does not affect child"`.

Here is that example alongside my new one, just to underscore their similarity:

```ruby
# Existing example:

it "adding parent setting does not affect child" do
  klass.setting :db, default: "sqlite"

  expect(subclass.settings).to eql(Set[:db])

  klass.setting :other

  expect(subclass.settings).to eql(Set[:db])
end

# New example:

specify "configuring the parent after subclassing does not copy the config to the child" do
  klass.setting :db

  subclass = Class.new(klass)

  object.config.db = "mariadb"

  expect(subclass.config.db).to be nil
end
```

The work inside this PR makes the settings shareable between the class and the `config` _inside a single class_, but because this behavior is contained to that single class only, it does _not_ change the overarching principal that settings and config are state, and that state is duplicated at the moment of subclassing.

In this way, it's not changing anything fundamental about dry-configurable's behavior.

I think if we wanted to make it possible for a parent class to have its config changed and all subclasses to receive that config, it'd require a lot of work and, based on the above, upend some meaningful assumptions baked into the current behavior.

Additionally, if some consumer of dry-configurable really needed this behavior, there'd be ways to make it possible by extending dry-configurable's behavior in the classes where it is used, rather than us modifying the library.

Given all of this, @flash-gordon (& @solnic), are you happy to go ahead with this PR and the behavior changes we've introduced here?

flash-gordon:
Sure, I agree. WRT how to propagate settings to child classes, it can be simple traversal via `.subclasses`/`.descendants`, I don't think there's need to change the design.

timriley:
> WRT how to propagate settings to child classes, it can be simple traversal via `.subclasses`/`.descendants`, I don't think there's need to change the design.

Oh yeah, I didn't think of descendent tracking — that could definitely be workable.

That said, do you think we _should_ actually do it? If, hypothetically, we were to ship 1.0 in the next couple weeks, do you think we should take the time to try this first?

flash-gordon:
My opinion is not propagating by default 💯 In the future, we could have an option for `setting`, that's it. Nothing blocking the release 🚢

timriley:
I've tested this locally with:

✅ dry-system
✅ dry-schema
✅ dry-validation
✅ dry-container
✅ dry-effects
✅ dry-monitor
✅ dry-rails
✅ hanami-view
✅ hanami (minor fix required for tests because I made a wrong assumption re: `Dry::Inflector` mutability)

And we're good to go! I'm going to merge and release this as a new `0.[THIS_ONE].0` release.

PR #131, opened 2022-03-31
Allow Hash as a possible config value

ojab:
It's handy to do something like
```ruby
v = {key: '"-----BEGIN RSA PRIVATE KEY-----…', key_password: 'qwerty'}
# and then
setting :openssl_key, constructor: ->(v) {
  OpenSSL::PKey.read(*v.values_at(:key, :key_password))
}
```
to have decrypted key in the settings instead of two strings, but dry-c complains that `nil` doesn't have `#update`. I suppose there is an expectation that if we're assigning a Hash — value is defined by `setting`, we shouldn't expect that.

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -57,8 +57,11 @@ module Dry
       # @api public
       def update(values)
         values.each do |key, value|
-          case value
-          when Hash
+          if self[key].is_a?(self.class)
+            unless value.is_a?(Hash)
+              raise ArgumentError, "#{value.inspect} is not a valid setting value"
+            end
+
             self[key].update(value)
           else
             self[key] = value
--- a/spec/integration/dry/configurable/config_spec.rb
+++ b/spec/integration/dry/configurable/config_spec.rb
@@ -35,6 +35,27 @@ RSpec.describe Dry::Configurable::Config do
       expect(klass.config.db.user).to eql("jane")
       expect(klass.config.db.pass).to eql("supersecret")
     end
+
+    it "runs constructors" do
+      klass.setting :db do
+        setting :user, default: "root", constructor: ->(v) { v.upcase }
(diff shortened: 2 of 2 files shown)


solnic:
I'm not sure what the actual problem is. Could you rephrase the description somehow? ie why is it trying to call `update` on `nil`?

ojab:
```ruby
require 'dry-configurable'

class Xxx
  extend Dry::Configurable

  setting :xxx
end

Xxx.configure do |config|
  config.update(xxx: {})
end
```
leads to
```
…/dry-configurable-0.14.0/lib/dry/configurable/config.rb:62:in `block in update': undefined method `update' for nil:NilClass (NoMethodError)

            self[key].update(value)
                     ^^^^^^^
	from …/dry-configurable-0.14.0/lib/dry/configurable/config.rb:59:in `each'
```
because it checks if input value is a `Hash` and we're passing the hash, but `:xxx` is not a subconfig, but a plain value.

ojab:
OTOH now we can overwrite `Config` with a plain value :/ I'll fix it a bit later today.

ojab:
BTW, shouldn't `.is_a?(Hash)` be `.respond_to(:to_hash)` instead? It's a convention that `#to_hash` is explicit permission for an object to be represented as a `Hash`. 
```
irb(main):001:0> x = Object.new
=> #<Object:0x00007f250e9bf278>
irb(main):002:0> def x.to_hash; {x: 1, y: 2};end 
=> :to_hash
irb(main):003:0> {**x}
=> {:x=>1, :y=>2}
```

ojab:
Added `ArgumentError` on assigning non-Hash to the setting.

timriley:
This all makes sense and looks good to me! I'll update the is_a?(Hash) check to use `to_hash` in a follow-up PR, then will cut a release.

timriley:
Released in 0.15.0: https://github.com/dry-rb/dry-configurable/releases/tag/v0.15.0

PR #136, opened 2022-08-06
Support for custom config classes

solnic:


solnic:
@timriley bump :)

timriley:
@solnic This is looking good to me so far! Feels like it'd be worth following through in this PR to support custom config classes on nested settings, because without that, this single "outer" custom config class may be of fairly limited value (i.e. I know we couldn't port Hanami::Action's configuration to vanilla dry-configurable at that point; we'd need the custom nested config classes too).

solnic:
@timriley you mean something like this:

```ruby
setting :db do
  setting :host, config: MyHostConfig do
    # ...
  end
end
```

??

timriley:
@solnic yep, exactly this!

solnic:
@timriley OK I'll make it possible in this PR

solnic:
@timriley I called the option `config_class` to avoid confusion with the actual config instance. Looks like things are working 🙂

timriley:
@solnic Looks great! And I agree on `config_class` being a clearer name.

Let's get this merged! Huge improvement for dry-configurable! 🎉 

Once this is merged I'll update #138 to fit with the new structure.

timriley (approved):

PR #138, opened 2022-08-27
Separate setting definitions from config values

timriley:
## Overview

In this PR we separate setting _definitions_ (i.e. the objects you create when you call `setting :my_setting`) from _config_, which now maintains its values in a separate hash, and stores only a reference to the setting definitions.

This allows settings objects to be shared across all config classes, which, along with some other changes here, reduces memory usage considerably.

According `benchmarks/memory_profile.rb` for 10,000 subclasses, this branch now allocates only 9.5 MB, down from 21.9 MB on the main branch.

## User-facing changes

Unless you're an advanced dry-configurable user here, and reaching deeply into our `Setting` objects, the user-facing changes here are minimal! Ordinary usages of dry-configurable will continue to work as before.

The one real behavioural change is that for instance-level configurable, new settings defined make themselves available even on previously initialized objects. This is in fact an incidental change to the rest of this PR and I could choose to wind that back, too, if we wanted.

## Implementation details

- `Setting` moves to being a strict _definition_ only. It no longer plays double duty as a realisation of an associated value within a config object. Specifically, it no longer has an `input` or `value`. Because of this, it no longer needs to be finalized or return a pristine copy or other such config-related responsibilities.
- `Setting::Nested` and `Setting#nested` have been removed. These no longer seemed necessary, and we now just provide a simple array of `children:` when initialising the setting.
- In turn `Settings` can lose config-related responsibilities such as `#finalize!` and `#pristine`.
- I renamed the internal `Settings` collection from `elements` to `settings` since it feels nicer.
- At the same time, this internal collection is no longer a `Concurrent::Map`, instead just a regular hash. This netted us some decent memory savings. **I am keen to hear feedback about this.** Why did it need to be a concurrent map in the first place? Are we worried e.g. about settings being defined in concurrent threads? What material downside would there be from this being a hash? And how does this compare with how we do things in e.g. `Dry::Core::ClassAttributes`, where there's no specific thread safety measures. Is calling `setting` here really all that different to `defines` from `ClassAttributes`?
- `Config` is now initialized with a `settings` instance, which it uses to determine which values it should be storing. These values are kept internally in a simple name=>value hash.
- When configurable classes are inherited, we create a dup of the `_settings` collection to the subclass. This dups its internal `settings` array only, so we don't copy all of the settings themselves, thereby preserving memory.
- Likewise, we create a copy of the `config` into the subclass using a new `Config#dup_for_settings` method, passing the new subclass settings collection to ensure the config object sees any new settings defined in the subclass. When we `#dup_for_settings`, we also dup any cloneable values that the config holds (i.e. values likely to be mutated and therefore ideally not shared), but we do not dup any of the other simple values, since they are safer to be shared.

## Next steps

In the shortest term, I'm going to let @solnic merge #136 when he is ready, and then I'll make sure I update this PR to work with those changes. I expect it should only need minimal adjustment, mostly just moving of code.

I also expect to follow this up in the next week with another proposal for further reducing dry-configurable memory usage, but this PR is already a good step in a positive direction, and I think it also helpfully clarifies the code at the same time.


timriley:
As of https://github.com/dry-rb/dry-configurable/commit/cb0a4919342d829482143c6832fec79648e41eec, here's the memory benchmarks:

For 100 subclasses:

```
        configurable   679.348k memsize (    77.844k retained)
                         5.044k objects (   443.000  retained)
                        50.000  strings (    50.000  retained)
configurable (nested)
                       229.568k memsize (     0.000  retained)
                         2.287k objects (     0.000  retained)
                        10.000  strings (     0.000  retained)

Comparison:
ordinary inheritance:      17776 allocated
configurable (nested):     229568 allocated - 12.91x more
        configurable:     679348 allocated - 38.22x more
```

vs main branch:

```
Comparison:
ordinary inheritance:      17776 allocated
configurable (nested):     530312 allocated - 29.83x more
        configurable:     742794 allocated - 41.79x more
```

For 10,000 subclasses:

```
        configurable    15.648M memsize (    77.844k retained)
                       143.644k objects (   443.000  retained)
                        50.000  strings (    50.000  retained)
configurable (nested)
                        22.168M memsize (     0.000  retained)
                       220.087k objects (     0.000  retained)
                        10.000  strings (     0.000  retained)

Comparison:
ordinary inheritance:    1760176 allocated
        configurable:   15648148 allocated - 8.89x more
configurable (nested):   22167968 allocated - 12.59x more
```

vs main branch:

```
Comparison:
ordinary inheritance:    1760176 allocated
        configurable:   21889194 allocated - 12.44x more
configurable (nested):   52247912 allocated - 29.68x more
```

So for 10,000 subclasses, this is a reasonable improvement:

- For top-level settings only, 8.89x our baseline vs 12.44x in the main branch
- For nested settings, 12.59x our baseline vs 29.68x in the main branch(!)

timriley:
A clearer comparison. For 10,000 subclasses with this branch:

```
Total allocated: 15647852 bytes (143651 objects)
Total retained:  7761316 bytes (50471 objects)

allocated memory by gem
   6280622  concurrent-ruby-1.1.10
   4801968  other
   4097136  dry-configurable/lib
    460501  zeitwerk-2.6.0
      7625  dry-core-8f5ca0ff8a19
```

And on the main branch:

```
Total allocated: 21888898 bytes (243654 objects)
Total retained:  16482777 bytes (180478 objects)

allocated memory by gem
   9000550  concurrent-ruby-1.1.10
   7618614  dry-configurable/lib
   4801968  other
    460141  zeitwerk-2.6.0
      7625  dry-core-8f5ca0ff8a19
```

That's 15MB allocated now, vs 21.8MB before, a 31.2% reduction.

timriley:
Memory impact of this PR on the Hanami r10k benchmark:

memory-dry-configurable-pr-138

Versus with the current release of dry-configurable:

memory-current

timriley:
14 days ago on this branch:

```
Total allocated: 15647852 bytes (143651 objects)
Total retained:  7761316 bytes (50471 objects)
```

Now:

```
Total allocated: 9511254 bytes (61150 objects)
Total retained:  7702930 bytes (50148 objects)
```

So this is 15.65MB allocated now down to 9.5MB (vs 21.89MB on the main branch).

One of the bigger gains here came from me switching to an ordinary Hash instead of a Concurrent::Map for `Settings#settings`. Are we particularly concerned about thread safety for adding to the setting definitions? I don't see how this is any different e.g. from Dry::Core::ClassAttributes, which just writes directly to class-level ivars when defining attributes.

lib/dry/configurable/class_methods.rb
```diff
@@ -11,8 +11,16 @@ module ClassMethods
...
 
-        subclass.instance_variable_set("@_settings", _settings.dup)
-        subclass.instance_variable_set("@_config", config.dup) if respond_to?(:config)
+        new_settings = _settings.dup
+
+        subclass.instance_variable_set(:"@_settings", new_settings)
+
+        # Only classes **extending** Dry::Configurable have class-level config. When
+        # Dry::Configurable is **included**, the class-level config method is undefined because it
+        # resides at the instance-level instead (see `Configurable.included`).
```
timriley, line 20: I left this note here because I remember being confused in the past about why we may not respond to `config` despite it being defined right here in this module. Hopefully this helps alleviate any confusion from readers in the future.

lib/dry/configurable/class_methods.rb
```diff
@@ -47,7 +55,7 @@ def setting(*args, **options, &block)
       #
       # @api public
       def settings
-        @settings ||= Set[*_settings.map(&:name)]
+        Set[*_settings.map(&:name)]
```
timriley, line 58: We don't want to memoize this because it'll capture stale data if called before all settings are defined.

lib/dry/configurable/compiler.rb
```diff
@@ -29,7 +29,9 @@ def visit_setting(node)
       # @api private
       def visit_nested(node)
         parent, children = node
-        visit(parent).nested(call(children))
+        name, opts = parent[1]
+
+        Setting.new(name, **opts, children: Settings.new(call(children)))
```
timriley, line 34: To save ourselves from creating a throwaway object (i.e. a setting that we then have to recreate extend with the children), we build the `Setting` instance here directly rather than running its part of the AST back through the compiler.

lib/dry/configurable/config.rb
```diff
@@ -65,61 +81,80 @@ def update(values)
...
+        setting = _settings[setting_name]
 
         super unless setting
 
-        if setting.writer?(meth)
-          raise FrozenConfig, "Cannot modify frozen config" if frozen?
-
-          _settings << setting.with(input: args[0])
+        if name.end_with?("=")
+          self[setting_name] = args[0]
```
timriley, line 135: Clarified reading and writing of values here by moving all such logic into the `#[]` and `#[]=` methods, rather than it being spread between them and `#method_missing`.

timriley:
[Reduce memory usage of Hanami::Action instances](https://trello.com/c/ExtpYdCm/208-reduce-memory-usage-of-hanamiaction-instances)

lib/dry/configurable/class_methods.rb
solnic: @timriley I would remove this, we have `_settings` and this one should be exposed as `settings` IMO. This Set is mostly here for historical reasons, it was just always like that but I'm not sure if this is a valuable thing to keep.

solnic:
custom config support was merged to main but I'm not sure how it's going to work here given that crucial construction logic was moved to the default `Config` class?

timriley:
> custom config support was merged to main but I'm not sure how it's going to work here given that crucial construction logic was moved to the default `Config` class?

I've rebased over the main branch now (and tidied my commit history) and everything is working fine! (👋🏼 @solnic)

If you'd like to review it, the `Config` construction logic is in these places:

- `ClassMethods#__config_build__` (same as your recent PR)
- `Setting#to_value` (changed for this PR, this used to be in the now-removed `Setting::Nested` class)

Overall, this PR is ready for review and merge once we're all good with it.

lib/dry/configurable/class_methods.rb
```diff
@@ -12,8 +12,16 @@ def inherited(subclass)
...
-        subclass.instance_variable_set(:@_config, config.dup) if respond_to?(:config)
+
+        new_settings = _settings.dup
+        subclass.instance_variable_set(:@_settings, new_settings)
+
+        # Only classes **extending** Dry::Configurable have class-level config. When
+        # Dry::Configurable is **included**, the class-level config method is undefined because it
+        # resides at the instance-level instead (see `Configurable.included`).
+        if respond_to?(:config)
+          subclass.instance_variable_set(:@config, config.dup_for_settings(new_settings))
```
timriley, line 23: Interestingly, until now we had been setting :@_config, when the actual memoized instance variable (used further below in the `#config` method) is `@config`. Setting this to the memoized ivar is in fact the desired behaviour here, so I think before now things must have been working accidentally.

lib/dry/configurable/dsl.rb
```diff
@@ -8,16 +8,12 @@ module Configurable
     class DSL
       VALID_NAME = /\A[a-z_]\w*\z/i.freeze
 
-      # @api private
```
timriley, line 11: This whole class is marked as `# @api private` so these lines aren't needed.

lib/dry/configurable/class_methods.rb
solnic: @timriley it'd be good to decide on just one approach for naming private ivars, in hanami I noticed we have `_foo`, in dry-rb we have `__foo__`. I'm just *mentioning* it here for future consideration.

lib/dry/configurable/class_methods.rb
solnic: @timriley the only reason why I introduced `_settings` was because I didn't want to break backward compatibility by changing `settings` from a set of symbol names to an array of setting objects.

lib/dry/configurable/config.rb
```diff
@@ -65,61 +81,80 @@ def update(values)
...
+      end
+
+      # Returns config values as a hash, with nested values also converted from {Config} instances
+      # into hashes.
+      #
+      # @return [Hash]
+      #
+      # @api public
+      def to_h
+        values.map { |key, value| [key, value.is_a?(self.class) ? value.to_h : value] }.to_h
```
solnic, line 105:
```suggestion
        values.to_h { |key, value| [key, value.is_a?(self.class) ? value.to_h : value] }
```

lib/dry/configurable/config.rb
```diff
@@ -65,61 +81,80 @@ def update(values)
...
+        _values.each_with_object({}) { |(key, val), dup_hsh|
+          dup_hsh[key] = _settings[key].cloneable? ? val.dup : val
+        }
       end
 
-      # @api private
       def initialize_copy(source)
         super
-        @_settings = source._settings.dup
+        @_values = source.send(:dup_values)
```
solnic, line 157:
```suggestion
        @_values = source.__send__(:dup_values)
```

lib/dry/configurable/setting.rb
```diff
@@ -19,106 +19,26 @@ class Setting
...
+      attr_reader :children
 
       # @api private
       def self.cloneable_value?(value)
         CLONEABLE_VALUE_TYPES.any? { |type| value.is_a?(type) }
       end
 
       # @api private
-      def initialize(name, input: Undefined, default: Undefined, **options)
+      def initialize(name, default: Undefined, children: nil, **options)
```
solnic, line 37: I'd do `children: EMPTY_ARRAY` and then down below instead of `!!children` it'd be `children.empty?`

lib/dry/configurable/settings.rb
```diff
@@ -38,37 +36,18 @@ def key?(name)
...
-      # @api private
-      def pristine
-        self.class.new(map(&:pristine))
-      end
-
-      # @api private
-      def finalize!(freeze_values: false)
-        each { |element| element.finalize!(freeze_values: freeze_values) }
-        freeze
+        settings.values.each(&block)
```
solnic, line 44:
```suggestion
        settings.each_value(&block)
```

solnic (approved):
Looking good overall, I left some suggestions that may improve performance a bit.

timriley:
Thanks @solnic! I'll incorporate all of your suggestions :)

And just to make sure, are you and @flash-gordon both fine with the replacement of the `Concurrent::Map` with an ordinary `Hash` for containing the setting definitions (inside `Settings`)?

jodosha (approved):

solnic:
> And just to make sure, are you and @flash-gordon both fine with the replacement of the Concurrent::Map with an ordinary Hash for containing the setting definitions (inside Settings)?

@timriley yes this is fine. If thread-safety is an issue here, some external mechanism should be incorporated to ensure the safety. Doing so at each library level turned out to be a bad idea, as we've learned 🙂

flash-gordon:
> And just to make sure, are you and @flash-gordon both fine with the replacement of the Concurrent::Map with an ordinary Hash for containing the setting definitions (inside Settings)?

I don't think this can lead to problems, practically speaking. Your reasoning is the same as mine, why would you have an environment where you define settings concurrently? AFAIK, `require` is thread-safe. There are possible other use cases but I agree with @solnic that they should be synchronized externally.

timriley:
Thank you all! Merging this one now.

ojab (commented):
TBH no idea how this supposed to be used after reading the specs.

timriley:
> TBH no idea how this supposed to be used after reading the specs.

@ojab What in particular are you looking to find out?

That the specs didn't change much indicates that overall usage remains the same, especially  if you're just using `setting` and `config`.

The biggest difference is that `_settings` now returns an array of `Dry::Configurable::Setting` objects that contain the details of the setting _definition_ only; the details about the configured _values_ of the settings are kept in `Dry::Configurable::Config` now, separated from the `Setting` instances.

ojab:
Thanks, I was trying to understand if there should be any user-visible changes, now I see that it's the internal change.

PR #140, opened 2022-09-17
Share config and reduce memory usage by writing values in `configure` blocks only

timriley:
This PR adopts a “copy in write” approach for handling `config` between classes and their subclasses, with a view to reducing memory usage as much as possible.

The `config` object is now completely _shared_ between a parent and its subclasses.

From there, when a subclass is configured, the config is copied in place as required, while still sharing as many internal _values_ as possible with the parent class; only the _changed_ internal values are themselves copied.

The way we achieve this is by having all mutations of `config` happen inside our `configure` blocks, just like we've supported since the beginning:

```ruby
configure do |config|
  config.foo = "bar"
  config.baz[:fred] = "barney"
end
```

The difference here is that this is now the _only_ places these mutations are supported: all mutative methods have moved from `Config` into a new `ConfigForUpdate` config wrapper class, which provides the object yielded to the `configure` block.

This `ConfigForUpdate` instance can track which values are accessed and changed and allows the `#configure` methods (the original one in our `Methods` mixing as well as a new one on `Config`) to determine whether to (a) copy the config object in the first place, and (b) which particular updated values to apply to that object.

This approach works for both simple values (which are assigned directly, e.g. strings) as well as mutable values that are updated in place (e.g. arrays, hashes, sets, or any other custom config values that users may provide).

This is a significant API change (you can no longer just call `config.foo = "bar"` directly in a class body), but I believe it’s worthwhile for the benefit of having dry-configurable to perform as efficiently as possible, and for our users to feel free to bring as many settings into their configuration as they like, without having to worry about undue memory usage in the case of many subclasses.

I've also added a new `benchmarks/memory_profile.rb` to analyse our memory usage. It's also a convenient way to share the improvements from this PR!

Let me walk you through two relatively common cases:

- Many subclasses that do not change any config from their parent class (`TIMES=10000 PROFILE=inherit_only bundle exec ruby benchmarks/memory_profile.rb`)
- Many subclasses that change only one or two config values from their parent class (`TIMES=10000 PROFILE=inherit_and_configure bundle exec ruby benchmarks/memory_profile.rb`)

For the first case (subclasses not changing config), this PR brings our 

| | current | new ✨ |
|--|--|--|
| allocated (MB) | 9.53 | 4.99 |
| retained (MB) | 7.71 | 4.83 |

For the second case (subclasses changing one config value):

| | current | new ✨ |
|--|--|--|
| allocated (MB) | 13.61 | 16.03 |
| retained (MB) | 9.39 | 7.31 |

If we dig into the "allocated objects by class" for that first case, you can see just how much we've optimised the allocations:

```
allocated objects by class
     10011  Class
      1228  String
        75  Array
        41  Hash
        20  Proc
        10  File
         3  Dry::Configurable::Setting
         3  Dry::Core::Equalizer
         3  Module
         1  Dry::Configurable::Compiler
         1  Dry::Configurable::Config
         1  Dry::Configurable::DSL
         1  Dry::Configurable::Extension
         1  Dry::Configurable::Settings
         1  Regexp
         1  Symbol
```

We may have 10,000 subclasses, but the `Dry::Configurable` objects are in the single digits only!

The one downside to this overall approach is that we do have to allocate an extra object during configuration, the `ConfigForUpdate` instance. This is why in the second memory profile case we actually end up allocating more memory than the main branch. Those objects are quickly GCed, however, and our overall retained memory footprint is lower. Do you think this is a reasonable tradeoff?

## Release plan

This is breaking change for dry-configurable and I propose that we do not offer any kind of graceful deprecation of the existing behavior. The reason for this is that the new behavior (no longer copying config on every inheritance) may cause insidious bugs for people who're currently relying on the side effects of this. It's better off they see a hard error and update their programs to the new API (which is very easy, it's just wrapping a `configure do` around the existing code).

Preparing for this to be released will also mean updating all existing ecosystem gems (dry-rb and Hanami) to ensure they wrap all config mutations in a `configure` block. I'll do this in a round of PRs after merging this and before releasing it. The changes in these gems will be both forward and backward compatible since all the existing releases of dry-configurable already support the `configure` API, and that will be unchanged in the new release.

Related PRs:

- https://github.com/dry-rb/dry-schema/pull/431
- https://github.com/dry-rb/dry-validation/pull/714
- https://github.com/dry-rb/dry-container/pull/88
- https://github.com/dry-rb/dry-system/pull/248
- No dry-monitor PR needed (it uses dry-configurable but no lib or test files actually change config

## Ready for 1.0?

Does this change mean we're ready for 1.0? My feeling is probably yes! Perhaps one possible way of releasing this change in a way that doesn't surprise existing users is to version it as 1.0.0.alpha1, that way any Gemfiles with version specs like `"~> 0.x"` will not pick it up.


timriley:
I made perhaps a more realistic benchmark, which does this:

```ruby
  klass = Class.new do
    extend Dry::Configurable

    setting :name
    setting :array_setting, default: %w[jane alice jen]
    setting :hash_setting, default: {"abcd" => "efgh"}
  end

  times.times do |i|
    subclass  = Class.new(klass) do
      configure do |config|
        config.name = "class #{i}"
      end
    end

    RETAIN << subclass
  end
```

Three settings, a couple with mutable values, then a bunch of subclasses, each of which tweaks one config value.

For this benchmark, we get the following:

| Branch                           | Allocated (MB) | Retained (MB) |
| -------------------------------- | ------------ | ----------- |
| `main`                             | 51.03        | 34.81       |
| `separate-setting-definitions`     | 13.61        | 9.39        |
| `write-settings-in-configure-only` | 14.73        | 7.31        |

So this branch is yet another improvement on the overall retained memory. It allocates _slightly_ more probably due to the transient `Config::UpdatingConfig` (name pending) classes that we're creating to track which values are changed inside `configure` blocks, but I think that is probably a fair tradeoff given configuration will only happen once.

lib/dry/configurable/config_for_update.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require "dry/core/constants"
+
+module Dry
+  module Configurable
+    # @api private
+    class ConfigForUpdate
```
solnic, line 8: Maybe `Config::Transient` would be better, it would look nicer and indicate it's private nature better as a nested class.

solnic (approved):

lib/dry/configurable/config_for_update.rb
timriley: Yeah, this is a good suggestion, thanks! I took the idea and ran with it a little.

`Transient`, while definitely a good descriptor for the lifecycle of the object, feel a little too generic to me — it's not clear from the name what the transient object is _used for_.

So I changed it to `Config::ToConfigure`, and changed the method returning this to be `Config#to_configure`. There's a nice symmetry there and hopefully this reinforces that this transient object is intended to live only over the lifecycle of the `#configure` method.

timriley:
Apart from the rename I just noted above, I've also updated this PR with a tweak that ensure we don't needlessly copy and reassign `@config` if a configurable class or object has `configure` run multiple times on it. Now the `@config` is copied/reassigned on first configure, and then reused in place for subsequent configures.

solnic:
> From the perspective of a user of dry-schema, this feels like a breaking change to me (she could previously mutate any exposed config without having to go through a configure block), so I think this probably warrants a 2.0 release of dry-schema (and any other dry-rb gem in this same situation).

Right, I pointed it out in the dry-schema PR too. So, it'd be good to preserve original behavior in dry-schema. It would not be great to bump schema/validation to 2.0 just because of this change. Alternatively, maybe it's just time to finally make the type spec change (making them mandatory) and push 2.0 along with the `configure` change?

timriley:
At this point I think it'd be good if we could just take whatever reasonable measures can see us get through our remaining to do list before Hanami 2. So for me personally, I'd actually feel comfortable bumping to 2.0 even just for this change. We've had a pretty good run on 1.x already, after all!

However, if you could have the type spec change ready to go say in the next two weeks or so (not sure how big this is, do let me know if that's unreasonable!), then I'd be fully on board for us rolling these two things together into 2.0 :)

solnic:
> However, if you could have the type spec change ready to go say in the next two weeks or so

I can and I'll do it. It's been too long already. I think it's important to do especially that with Hanami 2.0 release, more folks will be exposed to dry-schema/validation.

timriley:
@solnic Wonderful! Thank you!

timriley:
We've decided to close this PR since it would involve too much disruption across our ecosystem as well as broader users of dry-configurable.

PR #141, opened 2022-09-22
Add extension option to support `Undefined` as default value

### LINKED ISSUE #108: Leave setting values as `Undefined` rather than coercing to `nil` for un-set settings
This is something @solnic mentioned a while ago as a change he'd like to see for 1.0.0. Opening this so we can have the discussion as whether this is something we still want.

timriley:
If you include or extend dry-configurable like so:

```ruby
extend Dry::Configurable(default_undefined: true)
```

Then `Undefined` will become the default value for all settings instead of `nil` (when no setting-specific default is provided).

When a setting has no value and `Undefined` is configured as the default, then the constructor will not be called.

See my initial spec to demonstrate all of this in a nutshell:

```ruby
it "support Undefined as a default" do
  klass = Class.new do
    extend Dry::Configurable(default_undefined: true)

    setting :foo
    setting :bar, constructor: -> v { v.upcase }
    setting :baz, default: "baz", constructor: -> v { v.upcase }
  end

  expect(klass.config.foo).to be(Dry::Configurable::Undefined)
  expect(klass.config.bar).to be(Dry::Configurable::Undefined)
  expect(klass.config.baz).to eq "BAZ"
end
```

TBH, I think this PR is basically ready, apart from fleshing out that spec a little bit. I would love feedback at this point!

Resolves #108.


lib/dry/configurable/dsl.rb
```diff
@@ -26,107 +26,16 @@ def initialize(**options, &block)
...
-              "Provide a `constructor:` keyword argument instead",
-              tag: "dry-configurable",
-              uplevel: 2
-            )
-          end
-
-          options = options.merge(constructor: block)
-          block = nil
-        end
-
```
solnic, line 126: Oh boy I am so happy to see this go away

solnic (approved):
🚀 ❤️

lib/dry/configurable/dsl.rb
timriley: Haha, you and me both

flash-gordon (approved):
LGTM 👍

PR #143, opened 2022-10-04
Add Config#configured?

timriley:
Stop capturing all values on read (capture mutable values only in this case) and use the internal `_values` hash to determine whether a value has been configured.

This is useful at least for dry-system, which has had to implement [its own version of this method](https://github.com/dry-rb/dry-system/blob/2c4b246fe6e35011be2214ba6575e9a1655c0e0f/lib/dry/system/config/component_dir.rb#L262-L272), so is likely also useful for others, too.

The reason I'm making this change now is because this dry-system implementation depended on the internal structures of the `Config` object, which were changed in #138, thereby breaking dry-system. It's better we make this feature part of dry-configurable's public API at this point.

This change does impose a slight performance cost at config value read time, in that default values may need to be re-run through a setting's constructor (if defined), but I expect this would be minimal in ordinary use. If this turns out to be problematic in practice, we could later cache these read-but-not-assigned values into a separate internal hash, so I'm confident to go ahead with this change. I've chosen not to do any additional optimisations for now in order to keep the implementation code simpler.

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -41,7 +41,13 @@ module Dry
           raise ArgumentError, "+#{name}+ is not a setting name"
         end
 
-        _values.fetch(name) { _values[name] = setting.to_value }
+        _values.fetch(name) {
+          # When reading values, only capture cloneable (i.e. mutable) values in local state, making
+          # it easier to determine which values have actually been changed vs just read
+          setting.to_value.tap { |value|
+            _values[name] = value if setting.cloneable?
+          }
+        }
       end
 
       # Set config value.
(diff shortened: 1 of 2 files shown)


solnic (approved):
🚀

timriley:
Thanks @solnic!

I'll go and merge this now, with a view to doing a new 0.x.0 minor release of dry-configurable on Friday or over the weekend, clearing the way for a new dry-system release and then a hanami-controller one to follow.

timriley:
That is to say: still definitely open to feedback on this over the next few days if anyone has any :)

flash-gordon:
Does this mean the shim we have for dry-configurable in dry-container should be updated too? 🤔

timriley:
> Does this mean the shim we have for dry-configurable in dry-container should be updated too? 🤔 

@flash-gordon good question! Honestly, that feels like a fools errand to me.

It feels reasonable for our shim to maintain just the very basic behavior for reading and writing config. Chasing dry-configurable continually as it evolves (and I have no doubt it still has some evolving to do) feels like a waste of effort (and at some point, we'd end up maintaining two nearly-full parallel implementations).

solnic:
Yeah the line has to be drawn somewhere, and it feels like what we have now is the line already

PR #145, opened 2022-10-09
Share settings with subclasses where possible

timriley:
When a configurable class is inherited, do not eagerly create a dup of the `_settings` collection, and instead share the same collection with the subclass. Then, if that subclass define new settings, create a new `_settings` collection in place (along with a new matching `config` object referring to the new settings).

This change improves memory usage for the (quite common) scenario of many subclasses inheriting from a single class that contains all `setting` definitions.

For our `benchmarks/memory_profile.rb` with 10,000 subclasses inheriting from a single parent class that defines 3 settings (`TIMES=10000 PROFILE=inherit_only bundle exec ruby benchmarks/memory_profile.rb`), this exhibits the following memory usage:

```
Total allocated: 5765254 bytes (31228 objects)
Total retained:  5626120 bytes (30165 objects)
```

For the main branch, this is currently:

```
Total allocated: 9524491 bytes (61220 objects)
Total retained:  7706168 bytes (50166 objects)
```

**This is a 39.5% decrease in allocated memory for a large collection of subclasses (9.52 MB down to 5.77 MB).**

This change does have a couple of quirks, however:

- If you're inheriting from a configurable class and then capturing the `subclass.config` to a variable, and _then_ defining a new setting after, then the new `subclass.config` will be different to the one just captured.

  I do feel this is a reasonable tradeoff given the memory usage improvements. We should encourage our users to see `.config` (and `._settings`) as properties of the configurable class with their own opaque lifecycles; they should be accessed directly off the configurable class whenever needed.
- There is one behavioural change, in that a setting class defined in a parent class will also be visible to any subclasses even if those subclasses were made before the time of setting definition. Previously, settings from parent classes were only visible if they were defined before subclassing.

Overall, since this change considerably improves memory usage for our most common kind of usage (one or two parent classes with setting definitions, many subclasses without), I think it would be worth finding a way to get this in.

### DIFF
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -19,9 +19,13 @@ module Dry
       attr_reader :settings
 
       # @api private
+      attr_reader :lookup_cache
+
+      # @api private
       def initialize(settings = EMPTY_ARRAY)
         @parent_settings = []
         @settings = settings.each_with_object({}) { |s, m| m[s.name] = s }
+        @lookup_cache = {}
       end
 
       # @api private
@@ -45,22 +49,15 @@ module Dry
 
       # @api private
       def [](name)
-        all_settings[name]
-      end
-
-      # @api private
-      def key?(name)
-        keys.include?(name)
-      end
-
-      # @api private
-      def keys
-        all_settings.keys
+        lookup_cache.fetch(name) {
+          lookup_cache[name] = all_settings[name]
+        }
       end
 
       # @api private
       def each(&block)
-        all_settings.each_value(&block)
+        parent_settings.each { |parent| parent.each_value(&block) }
+        settings.each_value(&block)
       end
 
       private
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -32,17 +32,23 @@ module Dry
       private def initialize_copy(source)
         @parent_settings = source.parent_settings.dup
         @settings = source.settings.dup
+        @lookup_cache = source.lookup_cache.dup
       end
 
       # @api private
       def dup_for_child
-        dup.tap do |child_settings|
-          child_settings.parent_settings << settings
-        end
+        dup.tap { |child| child.add_parent(settings) }
+      end
+
+      # @api private
+      def add_parent(parent)
+        # Use length for all_settings cache invalidation
+        parent_settings << [parent, parent.length]
       end
 
       # @api private
       def <<(setting)
+        @all_settings = nil # bit gross
         settings[setting.name] = setting
         self
       end
@@ -56,14 +62,27 @@ module Dry
 
       # @api private
       def each(&block)
-        parent_settings.each { |parent| parent.each_value(&block) }
-        settings.each_value(&block)
+        all_settings.each_value(&block)
       end
 
       private
 
       def all_settings
-        {**parent_settings.reduce({}, :merge), **settings}
+        if parent_settings.none?
+          settings
+        elsif @all_settings && parent_settings.all? { |(parent, cached_length)| parent.length == cached_length }
+          # Our cache is valid; no new parent settings have been defined since it was built
+          @all_settings
+        else
+          @all_settings = {}.tap { |hsh|
+            parent_settings.each_with_index do |(parent, _length), index|
+              parent_settings[index][1] = parent.length
+              hsh.merge!(parent)
+            end
+
+            hsh.merge!(settings)
+          }
+        end
       end
     end
   end
--- a/lib/dry/configurable/settings.rb
+++ b/lib/dry/configurable/settings.rb
@@ -6,16 +6,35 @@ module Dry
     #
     # @api private
     class Settings
-      include Dry::Equalizer(:settings)
+      include Dry::Equalizer(:parent_settings, :settings)
 
       include Enumerable
+
+      # @return [Array<Hash>]
+      #
+      # @api private
+      attr_reader :parent_settings
 
       # @api private
       attr_reader :settings
 
       # @api private
       def initialize(settings = EMPTY_ARRAY)
+        @parent_settings = []
         @settings = settings.each_with_object({}) { |s, m| m[s.name] = s }
+      end
+
+      # @api private
+      private def initialize_copy(source)
+        @parent_settings = source.parent_settings.dup
+        @settings = source.settings.dup
+      end
+
+      # @api private
+      def dup_for_child
+        dup.tap do |child_settings|
+          child_settings.parent_settings << settings
+        end
       end
 
       # @api private
@@ -26,7 +45,7 @@ module Dry
 
       # @api private
       def [](name)
-        settings[name]
+        all_settings[name]
       end
 
       # @api private
@@ -36,18 +55,18 @@ module Dry
 
       # @api private
       def keys
-        settings.keys
+        all_settings.keys
       end
 
       # @api private
       def each(&block)
-        settings.each_value(&block)
+        all_settings.each_value(&block)
       end
 
       private
 
-      def initialize_copy(source)
-        @settings = source.settings.dup
+      def all_settings
+        {**parent_settings.reduce({}, :merge), **settings}
       end
     end
   end
--- a/lib/dry/configurable/class_methods.rb
+++ b/lib/dry/configurable/class_methods.rb
@@ -13,14 +13,15 @@ module Dry
 
         subclass.instance_variable_set(:@__config_extension__, __config_extension__)
 
-        new_settings = _settings.dup
-        subclass.instance_variable_set(:@_settings, new_settings)
+        # Share settings with subclasses until they define their own additional settings (see
+        # `.setting` below).
+        subclass.instance_variable_set(:@_settings, _settings)
 
         # Only classes **extending** Dry::Configurable have class-level config. When
         # Dry::Configurable is **included**, the class-level config method is undefined because it
         # resides at the instance-level instead (see `Configurable.included`).
         if respond_to?(:config)
-          subclass.instance_variable_set(:@config, config.dup_for_settings(new_settings))
+          subclass.instance_variable_set(:@config, config.dup)
         end
       end
 
@@ -42,6 +43,13 @@ module Dry
       # @api public
       def setting(*args, **options, &block)
         setting = __config_dsl__.setting(*args, **options, &block)
+
+        # If we're sharing settings with our superclass, create our own copy (along with a matching
+        # config copy) at the time of first new setting definition.
+        if superclass.respond_to?(:_settings) && _settings.eql?(superclass._settings)
+          @_settings = _settings.dup
+          @config = config.dup_for_settings(_settings) if respond_to?(:config)
+        end
 
         _settings << setting
 
--- a/lib/dry/configurable/class_methods.rb
+++ b/lib/dry/configurable/class_methods.rb
@@ -47,7 +47,7 @@ module Dry
         # If we're sharing settings with our superclass, create our own copy (along with a matching
         # config copy) at the time of first new setting definition.
         if superclass.respond_to?(:_settings) && _settings.eql?(superclass._settings)
-          @_settings = _settings.dup
+          @_settings = _settings.dup_for_child
           @config = config.dup_for_settings(_settings) if respond_to?(:config)
         end
 
--- a/lib/dry/configurable/class_methods.rb
+++ b/lib/dry/configurable/class_methods.rb
@@ -17,8 +17,8 @@ module Dry
         # `.setting` below).
         subclass.instance_variable_set(:@_settings, _settings)
 
-        # Only classes **extending** Dry::Configurable have class-level config. When
-        # Dry::Configurable is **included**, the class-level config method is undefined because it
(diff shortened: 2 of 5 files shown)


lib/dry/configurable/settings.rb
```diff
@@ -36,18 +55,18 @@ def key?(name)
...
-        settings.each_value(&block)
+        all_settings.each_value(&block)
       end
 
       private
 
-      def initialize_copy(source)
-        @settings = source.settings.dup
+      def all_settings
+        {**parent_settings.reduce({}, :merge), **settings}
```
timriley, line 69: I do wonder about the performance cost of this... but I suspect it's minimal.

solnic (commented):
> (Now fixed!) There is one behavioural change, in that a setting defined in a parent class will also be visible to any subclasses, unless those subclasses happen to have defined their own settings already.

That's **a really** confusing behavior and I'm pretty sure it will bite people. I also want to raise (again) the question of how relevant these benchmarks are to the actual real-world usage. Where/when are we creating thousands of configurable classes?

timriley:
@solnic could you please double check my last commit and the associated specs? I've addressed this issue now, and I'm not sure whether you're commenting on the original behavior I outlined, or the behavior I made consistent through my last commit.

timriley:
@solnic I've updated the second dot point to reflect the actual behavioural change. It now reads:

> There is one behavioural change, in that a setting class defined in a parent class will also be visible to any subclasses even if those subclasses were made before the time of setting definition. Previously, settings from parent classes were only visible if they were defined before subclassing.

Previously, it read:

> (Now fixed!) There is one behavioural change, in that a setting defined in a parent class will also be visible to any subclasses, _unless_ those subclasses happen to have defined their own settings already.
> 
> This one I feel a little less comfortable with. I don't expect this to bite people in ordinary usage, but if one does get into this situation, it's surprising behaviour. Now we _might_ be able to work around this by having our `Settings` class retain a an array of "parent" `settings` arrays, and look through them for setting objects in addition to its own local `settings` array. This should result in new settings defined in superclass appearing in all subclasses regardless of whether they happen to have their own local setting definitions. I might try this in another commit now. **UPDATE:** I addressed this in 737d94b, and it works consistently in both cases now.

timriley:
Now as for the _actual_ behavioural change:

> There is one behavioural change, in that a setting class defined in a parent class will also be visible to subclasses even if those subclasses were created before the time of setting definition. Previously, settings from parent classes were only visible if they were defined before subclassing.

You could make the case that this:

```ruby
class Parent
  extend Dry::Configurable
end

class Child < Parent
end

class Parent
  setting :foo, default: "bar"
end

Child.config.foo # => "bar
```

Is effectively equivalent to this, and should therefore not be surprising:

```ruby
class Parent
end

class Child < Parent
end

class Parent
  def self.foo
    "bar"
  end
end

Child.foo # => "bar"
```

The one downside is that this "changes in parent classes are visible even in previously-created subclasses" behaviour _only_ applies to setting definitions, not to actual config values.

So if a config value is set in a parent class _after_ a subclass has been created, it will _not_ propagate to the subclass:

```ruby
class Parent
  extend Dry::Configurable
  setting :foo, default: "bar"
end

class Child < Parent
end

Parent.config.foo = "custom foo"
Child.config.foo # => "bar"
```

This particular behaviour has not changed from previously, and again, you could make the argument of a distinction between settings and config, with settings being "behaviour" and config being "state", which would hopefully render this behaviour on both aspects unsurprising.

Either way, I don't expect either of these things to actually bite people in practice, since it's much more likely that both settings and config will be defined and left alone on parent classes before subclasses are created.

timriley:
> I also want to raise (again) the question of how relevant these benchmarks are to the actual real-world usage. Where/when are we creating thousands of configurable classes?

Yeah, likely not often. But if it was going to happen anywhere, a library like dry-configurable could very well be part of the mix. A very large Hanami 2.0 app, for example, could have at least hundreds of action and view classes.

This is why I thought it'd be worth sharing this PR. I'm happy to take it or leave it. But perhaps it'd be worth looking at the realistic case of e.g. ~500 configurable classes, half actions, half views, and see what the impact is.

solnic:
> @solnic could you please double check my last commit and the associated specs? I've addressed this issue now, and I'm not sure whether you're commenting on the original behavior I outlined, or the behavior I made consistent through my last commit.

Ahh, I thought it's still an issue. I missed the "(Now fixed!)" part 😞 Sorry for the noise

> This is why I thought it'd be worth sharing this PR. I'm happy to take it or leave it. But perhaps it'd be worth looking at the realistic case of e.g. ~500 configurable classes, half actions, half views, and see what the impact is.

I think the best benchmark is going to be a real Hanami 2.0 app with a more realistic number of configurable classes. 10k is a ridiculous number, if there's an app out there with so many configurable classes, then it's either some crazy meta-programming that generates these classes or the app is a monstrosity. I'm all for optimization and lowering memory footprint but I also want to be sure we're targeting more common and realistic scenarios.

Here's something to think about: an almost empty Rails app I've got takes ~180MB, then I have ActionHero which has way more deps and actual custom code and it takes ~135MB 🤷

timriley:
So the largest Hanami 2 app I know has ~150 actions.

If I drop into hanami-controller and tweak its memory_profile_action.rb so it's just focused on the cost of subclassing from a shared base class, I get these results:

```
Total allocated: 254400 bytes (900 objects)
Total retained:  229200 bytes (750 objects)

allocated memory by gem
    182400  dry-configurable-0.16.0
     72000  other
```

(I included the allocated memory by gem to show that dry-configurable is pretty much the biggest third-party allocator of memory in this situation.)

If I then go into that codebase and have it use this particular branch of dry-configurable, then we see the following:

```
Total allocated: 84000 bytes (450 objects)
Total retained:  84000 bytes (450 objects)

allocated memory by gem
     72000  other
     12000  dry-configurable/lib
```

So this is a change from 254400 bytes (0.25 MB) allocated to 84000 bytes (0.08 MB), a reduction of 67%.

If we double the number of actions to 300, to approximate their corresponding view classes, this becomes a change from 508800 bytes (0.51 MB) to 168000 bytes (0.17 MB), another reduction of 67%.

These are small numbers overall, yes, but still meaningful reductions in the amount of memory dry-configurable is responsible for! In my opinion, this change would represent us reaching the best possible memory performance from dry-configurable in its current incarnation.

So at this point, my question here would be: are we happy with the subtle change in behaviour I described above (settings are shared across all subclasses no matter when they are defined, while configs are shared only if set before subclassing), along with the actual code changes in this PR? If that's the case, then I think we could merge this and draw a line under our performance work here.

If, however, any of us have doubts or misgivings about the changes here, then I think we could _also_ be happy just to close this off and still be satisfied with the work we've previously done to improve dry-configurable memory usage.

So: what do you think?

solnic:
@timriley ❤️ thanks for taking the extra time to check this! Yes I think it makes sense to have this merged. The change in behavior is actually an improvement (IIRC it used to work like that anyway).

solnic (approved):

timriley:
[One last pass at optimising dry-configurable](https://trello.com/c/UonNhzbI/240-one-last-pass-at-optimising-dry-configurable)

timriley:
Simpler benchmark doing a single `Config#to_h`:

```ruby
# frozen_string_literal: true

require "bundler/setup"

require "benchmark/ips"
require "dry/configurable"

class Base
  extend Dry::Configurable

  setting :foo
  setting :bar
  setting :baz, default: "baz"
  setting :qux, default: "qux", constructor: :upcase.to_proc

  config.bar = "bar"
end

class Inherited < Base
  setting :quux
end

def name(str)
  [ENV["BENCHMARK_NAME"], "#{str}"].compact.join("/")
end

Benchmark.ips do |x|
  x.warmup = 1
  x.time = 3

  x.report(name("base/to_h")) do
    Base.config.to_h
  end

  x.report(name("inherited/to_h")) do
    Inherited.config.to_h
  end

  x.save! ENV["SAVE_FILE"] if ENV["SAVE_FILE"]
  x.compare!
end
```

Results:

```
               main/base/to_h: 223775.6 i/s
          main/inherited/to_h: 176825.3 i/s - 1.27x  (± 0.00) slower
     share-settings/base/to_h: 176277.4 i/s - 1.27x  (± 0.00) slower
share-settings/inherited/to_h: 114801.0 i/s - 1.95x  (± 0.00) slower
```

timriley:
Actually a more striking difference if we benchmark back at 22ab1fa, which is before we merged #143:

```
            22ab1fa/base/to_h: 968883.0 i/s
       22ab1fa/inherited/to_h: 807447.9 i/s - 1.20x  (± 0.00) slower
               main/base/to_h: 223775.6 i/s - 4.33x  (± 0.00) slower
          main/inherited/to_h: 176825.3 i/s - 5.48x  (± 0.00) slower
     share-settings/base/to_h: 176277.4 i/s - 5.50x  (± 0.00) slower
share-settings/inherited/to_h: 114801.0 i/s - 8.44x  (± 0.00) slower
```

In #143 we changed `Config` so it doesn't capture non-clone able values into `@_values` on read, for the purposes of implementing `#configured?`. I might be able to improve on that by using simple `==` checks against setting defaults for `#configured?` instead, though, which might be enough for us to move forward with this PR.

timriley:
Now that I've checked in a benchmark to give us a good indication of overall config read performance (in #152), I've run this benchmark against this branch, with #149 (which I expect will merge soon) as baseline. Here are the results:

```
Comparison:
           base/base:   968961.3 i/s
            base/sub:   807698.1 i/s - 1.20x  (± 0.00) slower
 share-settings/base:   174239.5 i/s - 5.56x  (± 0.00) slower
  share-settings/sub:   112031.8 i/s - 8.65x  (± 0.00) slower
```

So for a subclass with settings definitions on two of its parents (the `/sub` entries above), this PR is 8.65x slower on reading config values.

This is no doubt because of the work we're doing in `Settings#all_settings`, which is called when reading each config key to determine whether the key corresponds to a defined setting:

```
def all_settings
  {**parent_settings.reduce({}, :merge), **settings}
end
```

I'm not sure if a nearly 10x perf hit is worth it for what we're otherwise achieving in this PR.

If we did feel that the behavioural and memory-related changes in this PR are worth keeping, we could look to improve the performance. We could e.g.

- Cache settings at the level of `Config` instances so we don't have to look them up through those parent_settings each time — this seems wasteful and would effectively undo any memory savings we got from this PR in the first place
- "Flatten" the settings within `Settings` instances, keeping a secondary hash that points directly to a setting for a given name, regardless of whether it came from a parent or not. This might be promising. It wouldn't add _too_ much complexity, and because we're sharing `settings` objects as much as possible, wouldn't take too much memory either.

I'm going to take a quick spike of the second option now, just so we can see what it looks like.

From there, I think we should make a quick decision about whether to accept or abandon this PR.

timriley:
Alright, in b658e06 I added the lookup cache to `Settings`, but I should note we can only use it when looking up individual settings. If we want to know about _all_ settings, we still have to loop through all parent settings because inside `Config` we have no way to tell if or when new settings are defined.

I was curious, so I made a simple benchmark for reading a single config value, and here the difference between this branch and our base:

```
Comparison:
            base/sub:  1745591.1 i/s
           base/base:  1741132.6 i/s - same-ish: difference falls within error
  share-settings/sub:  1653854.7 i/s - 1.06x  (± 0.00) slower
 share-settings/base:  1645729.7 i/s - 1.06x  (± 0.00) slower
```

timriley:
And now in 9c25345 I've taken this about as far as I think I can take it. I realised that I can use the `length` of the parent setting objects as a kind of ghetto cache invalidation key for when we want to loop over all settings. If we capture the length each parent settings object when we receive it, we can use that to know whether new settings have been defined within them or not, and on that basis, we can build up a cache for `Settings#all_settings`.

This gives us the following new benchmark for `Config#to_h`:

```
Comparison:
           base/base:   968961.3 i/s
 share-settings/base:   911647.4 i/s - 1.06x  (± 0.00) slower
            base/sub:   807698.1 i/s - 1.20x  (± 0.00) slower
  share-settings/sub:   724257.9 i/s - 1.34x  (± 0.00) slower
```

So I guess everything now boils down to this:

For the (admittedly meagre) memory savings from sharing `settings` objects, and the (admittedly nice) bonus behaviour of having new parent settings propagate down to child classes even if those child classes were already defined, are we prepared to accept:

- Marginally slower individual config reads (1.06x slower, per my previous comment)
- Marginally slower whole config reads (1.115x slower, comparing the two `/sub` entries above)
- The added complexity inside `Settings` due to the extra caching we need to have (n.b. I haven't measured the memory impact of this)
- The small quirk that `.config` on a subclass of a configurable object may change after you define a first setting at that level (IMO this is not a big deal overall, but it's worth remembering)

My overall feeling is that **NO**, these tradeoffs are not worth it. I think we benefit more from the conceptual and implementation simplicity in dry-configurable’s current state versus the memory wins we get here, and we should not merge this PR.

I'd be interested in a quick check-in from you all, @solnic, @flash-gordon, and also @jodosha since I know you were looking for even more memory wins from this library. As I showed in #144, and from the added complexity here, I think we've probably got things are good as we can reasonably make them, short of overhauling this whole thing (and even doing that wouldn't guarantee we get into any significantly better position).

PR #149, opened 2022-10-11
Restore Config value read performance

timriley:
Restore the performance of `Config` on value reads back to its state before #143.

Here we capture all values on read to `Config`'s internal `_values` hash. This means that after first read, all subsequent reads should be fast (This is how it was before #143).

Then to preserve the functionality of `#configured?` that we introduced in #143, we keep a separate `_configured` Set that tracks the names of values that have either been assigned, or mutable values that have been read. For the very small amount of memory (one set per config object, since most of the names, as symbols, should already be in memory) this is a worthwhile tradeoff for the best possible performance on read, which should reasonably be expected by our users to be as fast as possible.

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -130,7 +130,10 @@ module Dry
       #
       # @api public
       def values
-        _settings.to_h { |setting| [setting.name, self[setting.name]] }
+        # Ensure all settings are represented in values
+        _settings.each { |setting| self[setting.name] unless _values.key?(setting.name) }
+
+        _values
       end
 
       # Returns config values as a hash, with nested values also converted from {Config} instances
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "dry/core/constants"
+require "set"
 
 module Dry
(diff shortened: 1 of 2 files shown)


solnic (commented):
I'm missing something because it's hard for me to understand why this was needed. Given that we already have `_values` why we suddenly also need `_configured`? What's the difference between the two?

timriley:
`@_values` is the hash where the config values are kept, regardless of where they came from (i.e. via a setting default or explicit assignment). We keep them all in this hash to ensure that reads remain quick. This is how it was in previous releases of the library.

In #143 I changed how we populated `_values` so that it could be used to determine whether a user had explicitly assigned a value. But this had some negative performance ramifications. **This change was NOT a part of the memory optimisation effort; it was me adding what I believed to be a missing feature, based on places we'd been using dry-configurable.**

To restore config read performance, in this PR I'm letting `_values` be populated again on every read. This means we now need a separate place to track whether a user has explicitly assigned a config value, hence `_configured`.

timriley:
@solnic @flash-gordon Folks, this is ready to go. Since it is a fix to a performance regression I'm incline to get it in sooner rather than later and ship a patch release. Please let me know if you have any concerns :)

flash-gordon (approved):

PR #153, opened 2022-10-12
Rename `_settings` class method to `settings`

timriley:
This gives us a "neater" name for the second (after `config`) main way of interacting with configurable classes. It's also better match with the `setting` method we use for defining new settings in the first place.

This method was previously `_settings` because we had already had the `settings` name taken by the method that returns a set of setting names only. However, these names can be just as easily accessed via the `#name` attribute on individual settings themselves, and reducing our overall surface area here will help make the settings side of our API easier to understand and use.

Along with this change, I've made a range of methods `@api public` for the first time, on both `Settings` and `Setting`, since these are the objects returned by `.settings`. This is more a reflection of real usage than a big change, however, since I expect we would have had users already interacting with these objects given that the previous `._settings` method was itself already marked as `@api public`.

**This is a breaking change**, though easy to correct (change `.settings` in user code to `.settings.map(&:name)`.

### DIFF
--- a/lib/dry/configurable/class_methods.rb
+++ b/lib/dry/configurable/class_methods.rb
@@ -13,7 +13,7 @@ module Dry
 
         subclass.instance_variable_set(:@__config_extension__, __config_extension__)
 
-        new_settings = _settings.dup
+        new_settings = settings.dup
         subclass.instance_variable_set(:@_settings, new_settings)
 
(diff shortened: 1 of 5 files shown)


timriley:
[Rename _settings to settings in dry-configurable](https://trello.com/c/qBnJuHPE/241-rename-settings-to-settings-in-dry-configurable)

solnic (approved):

PR #155, opened 2022-10-13
Exit from `Config#finalize!` if already frozen

timriley:
This saves the work of attempting to finalize every nested config (or attempt to freeze again every value) if `finalize!` is called a second time.

This does make `finalize!` a one-shot operation, i.e. you can't call `finalize!` first, followed by `finalize!(freeze_values: true)`, but doing this would not be expected usage.

Also: rename error classes:

- `Dry::Configurable::AlreadyIncluded` -> `AlreadyIncludedError`
- `Dry::Configurable::FrozenConfig` -> `FrozenConfigError`

### DIFF
--- a/spec/integration/dry/configurable/config_spec.rb
+++ b/spec/integration/dry/configurable/config_spec.rb
@@ -261,6 +261,61 @@ RSpec.describe Dry::Configurable::Config do
     end
   end
 
+  describe "finalize!" do
+    it "can be finalized" do
+      klass.setting :db do
+        setting :user, default: "root".dup
+      end
+
+      klass.finalize!
+      klass.finalize! # second call becomes a no op
+
+      # It freezes the config and nested config classes
+      expect(klass.config).to be_frozen
+      expect(klass.config.db).to be_frozen
+
+      # But does not freeze values by default
+      expect(klass.config.db.user).not_to be_frozen
+      klass.config.db.user << "foo"
+      expect(klass.config.db.user).to eq("rootfoo")
+
+      # It does not allow configure block anymore
+      expect { klass.configure {} }.to raise_error(Dry::Configurable::FrozenConfigError)
+    end
+
+    it "can be finalized with freezing values" do
+      klass.setting :db do
+        setting :user, default: "root".dup
+      end
+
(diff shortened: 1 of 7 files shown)


solnic (approved):
Could you exercise this in a spec please?

timriley:
> Could you exercise this in a spec please?

@solnic Gladly!

I've done three changes here:

- 4a893c7, which adds the tests (and rearranges some a little bit)
- 681687d, which improves the frozen check used inside the `configure` method (previously it would not have raised the error when `Dry::Configurable` is extended
- 70c30c9, made our custom error class names conventional by adding "Error" on the end. Not directly required for the rest of this PR but it needed a tidy and I just wanted to get it sorted

spec/integration/dry/configurable/config_spec.rb
```diff
@@ -261,6 +261,61 @@
...
+      end
+
+      klass.finalize!
+      expect(klass.config).to be_frozen
+      expect(klass.config.db.user).not_to be_frozen
+
+      # Calling the second time does not do any work
+      klass.finalize!(freeze_values: true)
+      expect(klass.config.db.user).not_to be_frozen
+    end
```
timriley, line 316: This specifically is the test that shows the second finalize does no work

timriley: The tests above this were just moved over from the setting_spec.rb, because they didn't really make sense there.

spec/integration/dry/configurable/setting_spec.rb
```diff
@@ -398,6 +398,14 @@
       expect(object_2.config.after).to eq "defined after"
     end
 
+    it "is frozen when finalized" do
+      object.finalize!
+      object.finalize! # second call becomes a no op
+
+      expect(object).to be_frozen
+      expect(object.config).to be_frozen
+    end
```
timriley, line 407: This test just makes it clearer that calling finalize will freeze the object itself (as well as config) when `Dry::Configurable` is included.

timriley:
I'll leave this for another day if you wanted to give it a final look before I merge, @solnic :)

PR #159, opened 2022-11-15
Dunderscore mixed in ivar names to ensure no clashes with user-set vars

timriley:
We saw a clash in a dry-system provider, which includes `Dry::Configurable`, in which user-supplied code could set a `@config` ivar which suddenly magically became accessible via the `#config` method (instead of the actual dry-configurable config).


timriley:
[Prefix dry-system Provider ivars with _ to avoid user collisions](https://trello.com/c/bXssvifb/336-prefix-dry-system-provider-ivars-with-to-avoid-user-collisions)

solnic (approved):

jodosha (approved):

Issue #158, opened 2022-10-20
0.15.0 broke `config.update` with constuctor requiring nested value


ojab:
Bisected to https://github.com/dry-rb/dry-configurable/commit/5a5efd7d2e84fd6dc2515afa50cbf45a23ec24af (yeah, I see the commit author)

```ruby 
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'dry-configurable', '= 0.15.0'
end

module App
  extend Dry::Configurable

  setting :foo do
    setting :bar, constructor: ->(value) { value * 2 }
  end
end

App.configure do |config|
  config.update({'foo' => {'bar' => 1}})
end

puts App.config.foo.bar
```
with dry-c 0.14.0 it prints `2` as expected, with 0.15.0 it's
```
/tmp/d-c.rb:14:in `block (2 levels) in <module:App>': undefined method `*' for nil:NilClass (NoMethodError)

    setting :bar, constructor: ->(value) { value * 2 }
                                                 ^
	from …/lib/dry/configurable/setting.rb:89:in `value'
	from …/lib/dry/configurable/config.rb:39:in `[]'
```

I'll try to take a look, but not earlier than next week :(

ojab:
AFAIU it's caused by https://github.com/dry-rb/dry-configurable/blob/main/lib/dry/configurable/setting.rb#L78 and `Undefined` should be `default config value`, so we need to pass `default_undefined` there and from a quick glance I don't see an easy way.

PR #161, opened 2023-07-14
Allow nested settings to default to `Undefined`

emptyflask:
I noticed that when using `Dry::Configurable(default_undefined: true)`, nested settings still default to `nil`. This ought to fix that.

And I unfroze a regex at the request of rubocop.


emptyflask:
This might also fix https://github.com/dry-rb/dry-configurable/issues/158

solnic:
Thank you! <3

PR #164, opened 2024-06-17
Allow multiple includes of Dry::Configurable

timriley:
Instead of raising AlreadyIncludedError (now removed), update the method undef code in `.included` to handle already-undefined methods gracefully.

Allowing Dry::Configurable to be included multiple times across subclasses now allows for uses cases like a subclass adjusting various extension settings, e.g. `config_class`.

### DIFF
--- a/lib/dry/configurable/errors.rb
+++ b/lib/dry/configurable/errors.rb
@@ -5,8 +5,16 @@ module Dry
   #
   # @api public
   module Configurable
+    extend Dry::Core::Deprecations["dry-configurable"]
+
     Error = Class.new(::StandardError)
 
     FrozenConfigError = Class.new(Error)
+
+    AlreadyIncludedError = Class.new(Error)
+    deprecate_constant(
+      :AlreadyIncludedError,
+      message: "`include Dry::Configurable` more than once (if needed)"
+    )
   end
 end
--- a/lib/dry/configurable/errors.rb
+++ b/lib/dry/configurable/errors.rb
@@ -7,7 +7,6 @@ module Dry
   module Configurable
     Error = Class.new(::StandardError)
 
-    AlreadyIncludedError = Class.new(Error)
     FrozenConfigError = Class.new(Error)
   end
 end
--- a/lib/dry/configurable/extension.rb
+++ b/lib/dry/configurable/extension.rb
@@ -35,7 +35,7 @@ module Dry
 
           class << self
             undef :config if method_defined?(:config)
-            undef :configure  if method_defined?(:configure)
+            undef :configure if method_defined?(:configure)
           end
         end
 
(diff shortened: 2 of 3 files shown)


lib/dry/configurable/errors.rb
```diff
@@ -7,7 +7,6 @@ module Dry
   module Configurable
     Error = Class.new(::StandardError)
 
-    AlreadyIncludedError = Class.new(Error)
```
flash-gordon, line 10: Hard to tell, but it may be a breaking change if somebody references this constant in their code

lib/dry/configurable/extension.rb
```diff
@@ -36,8 +34,8 @@ def included(klass)
           prepend(Initializer)
```
flash-gordon, line 34: Now that these lines are reachable, things are more tricky. `include` and `prepend` work differently, `include` will add a module to the ancestors' chain only once, whereas `prepend` does it every time. We should keep this in mind

lib/dry/configurable/errors.rb
timriley: Ah yes, thanks for noting that. I'll put the class back in for now, along with a comment that notes it will be deprecated soon.

flash-gordon: btw we have a mechanism for deprecating constants in dry-core

timriley: Thanks again @flash-gordon, have addressed this in ca63b93 :)

timriley:
@flash-gordon FYI, this is the reason for this PR: [hanami/hanami@`86db090` (#1430)](https://github.com/hanami/hanami/pull/1430/commits/86db09019de42ee5a3fadc3334cbf866ed05183e)

`Dry::System::Provider::Source` already includes `Dry::Configurable`, providing a generic `config` object, which is fine in most circumstances. For the Hanami's `DB` provider source, however, we have fairly advanced config needs, so I want to provide a custom config API. By re-including `Dry::Configurable` and providing the `config_class:` option, we can now do that.

How does that feel to you?

I thought about adding a dedicated API to `Provider::Source` for specifying a custom config class, but that felt like overkill when I'm not sure how widely it will be needed, and our needs for the `DB` provider can be taken care of by a straightforward `include`. One downside of this approach is that it presumes knowledge of `Provider::Source`'s internals (i.e. that it uses `Dry::Configurable`), but given the close relationship between Hanami and dry-rb, I think we can manage this.

lib/dry/configurable/extension.rb
timriley: Good call. I'll look to expand the testing to cover a few more possible edge cases before I mark this ready for a final review.

timriley:
@flash-gordon I've expanded the tests to make sure nothing surprising happens with the multiple prepended `#initialize` methods. See the last commit above. It all looks good to me, things work as they should.

Given this, are you happy to see this be merged? Thanks again for checking!

flash-gordon (approved):

PR #167, opened 2026-05-06
Add Config#to_data to allow faster reads

cllns:
I explored doing this upstream in https://github.com/hanami/hanami-view/pull/276 but this is something we can, and should, add to dry-configurable.

This is similar to adding `reader: true` but that adds an attr_reader on the host itself, not the config. This returns a new `Data` object wrapping the settings (after finalize).

You have to cache it by object_id (or using Hash#cache_by_identity) in order to get all the benefits of it, in a library like hanami-view where classes are meant to be instantiated once and the instances re-used.

| Pattern | ns/op | vs `config.foo` |
|---|---:|---:|
| `cached.foo` (memoized `to_data`) | 38 | **13.6× faster** |
| `config.foo` (existing, method_missing) | 517 | 1× (baseline) |
| `config.to_data.foo` (fresh per call) | 838 | 1.6× slower |

### DIFF
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -153,6 +153,31 @@ module Dry
         _dry_equalizer_hash
       end
 
+      # Returns a frozen {Data} representation of the config's resolved setting values.
+      #
+      # The Data exposes each setting as a real method, sidestepping the {method_missing}
+      # dispatch on the read path. Intended for performance-sensitive code that reads the same
+      # config repeatedly (e.g. per-request render hot paths). A fresh Data is built per call,
+      # so consumers should memoize the result.
+      #
+      # Only available on a finalized config. Nested configs are converted recursively to nested
+      # Data instances. Values are captured by reference, so in-place mutation of a captured
+      # value remains visible through the Data; finalize with `freeze_values: true` if you want
+      # to prevent that.
+      #
+      # @return [Data]
+      #
+      # @raise [FrozenConfigError] if the config is not yet finalized
+      #
+      # @api public
+      def to_data
+        unless frozen?
+          raise FrozenConfigError, "config must be finalized before #to_data can be called"
+        end
+
+        _settings.data_class.new(**to_data_attrs)
+      end
+
       # @api public
       def finalize!(freeze_values: false)
         return self if frozen?
@@ -181,6 +206,12 @@ module Dry
       end
 
       private
+
+      def to_data_attrs
+        _values.each_with_object({}) do |(name, value), hsh|
+          hsh[name] = value.is_a?(self.class) ? value.to_data : value
+        end
+      end
 
       def method_missing(name, *args)
         setting_name = setting_name_from_method(name)
--- a/lib/dry/configurable/config.rb
+++ b/lib/dry/configurable/config.rb
@@ -165,6 +165,23 @@ module Dry
       # value remains visible through the Data; finalize with `freeze_values: true` if you want
       # to prevent that.
       #
+      # The returned Data uses Ruby's default structural `==`/`eql?`/`hash`. If you want to use
+      # it as an identity-based cache key (e.g. to skip walking N members on every lookup),
+      # memoize the reference yourself and either:
+      #
+      #   1. key it by `object_id` directly:
+      #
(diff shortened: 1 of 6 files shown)


lib/dry/configurable/config.rb
```diff
@@ -153,6 +153,48 @@ def hash
         _dry_equalizer_hash
       end
 
+      # Returns a frozen {Data} representation of the config's resolved setting values.
```
timriley:
```suggestion
      # Returns a frozen {Data} representation of the config's resolved values.
```

"setting values" feels a bit awkward as a term. we're in the config already, so just "values" feels fine here.

lib/dry/configurable/config.rb
```diff
@@ -153,6 +153,48 @@ def hash
...
+      #
+      # The Data exposes each setting as a real method, sidestepping the {method_missing}
+      # dispatch on the read path. Intended for performance-sensitive code that reads the same
+      # config repeatedly (e.g. per-request render hot paths). A fresh Data is built per call,
```
timriley:
```suggestion
      # config repeatedly (e.g. per-request render hot paths). A new object is returned per call,
```

Feels a bit clearer and more natural to read (you might need to re-wrap it, sorry)

lib/dry/configurable/config.rb
```diff
@@ -153,6 +153,48 @@ def hash
...
+      # structural equality available for general use.
+      #
+      # @return [Data]
+      #
+      # @raise [FrozenConfigError] if the config is not yet finalized
+      #
+      # @api public
+      def to_data
+        unless frozen?
+          raise FrozenConfigError, "config must be finalized before #to_data can be called"
```
timriley, line 192: Feels a bit strange at first to see us `FrozenConfigError` for both when the config _is_ and _is not_ frozen, but I also prefer not to create a ton of hyper-specific error classes, so I think this is ultimately the right decision :)

No need for a change here, just wanted to share my impression as I first read it.

lib/dry/configurable/dsl.rb
```diff
@@ -31,6 +35,12 @@ def setting(name, **options, &block)
           raise ArgumentError, "#{name} is not a valid setting name"
         end
 
+        if DATA_RESERVED_NAMES.include?(name.to_sym)
+          raise ArgumentError,
+                "#{name.inspect} is not a valid setting name: it conflicts with a Data " \
+                "instance method, which would break Config#to_data"
+        end
```
timriley, line 42: This is quite a lot of names we're reserving, but I think it's unlikely people will want to take any of them:

```ruby
irb(main):003> puts Data.instance_methods
pretty_print
deconstruct_keys
deconstruct
pretty_print_cycle
==
hash
members
inspect
with
to_s
eql?
to_h
pretty_print_inspect
pretty_print_instance_variables
singleton_class
dup
itself
methods
singleton_methods
protected_methods
private_methods
public_methods
instance_variables
instance_variable_get
instance_variable_set
instance_variable_defined?
remove_instance_variable
instance_of?
kind_of?
is_a?
display
frozen?
class
then
yield_self
tap
public_send
extend
clone
<=>
===
!~
nil?
method
respond_to?
public_method
singleton_method
define_singleton_method
freeze
object_id
send
to_enum
enum_for
pretty_inspect
equal?
!
__send__
!=
__id__
instance_eval
instance_exec
```

So I think we're safe to go ahead with this decision. We can always reconsider our approach based on user feedback in the future.

timriley (approved):
This is a nice enhancement, thank you @cllns!

Left you a few small thoughts. No meaningful changes needed.

Happy for you to merge whenever you're ready. Can you please add a CHANGELOG entry before doing so?

lib/dry/configurable/dsl.rb
cllns: Yea, I think it's better to error out explicitly in those cases anyway, rather than let users override them (which Data.define/new let you do) and discover some downstream error with missing an expected method.

cllns:
Addressed all feedback and updated CHANGELOG. Will merge tomorrow to give people a chance to read over it if they have concerns, or feel free to merge whenever you'd like @timriley

